PR #203, opened 2024-04-20
Add an option to specify the core_id when initializing I2C driver

sodank808:
In situations where all non-reserved core 0 interrupt slots are taken, it is helpful to be able to move the I2C interrupts to another core if available.


finger563 (changes requested):
This code doesn't do what the comment on the config argument says - and it also doesn't handle the case where a value less than -1 is inadvertently passed. I'd recommend at least handling that case and updating the comment.

finger563 (approved):

PR #214, opened 2024-04-27
Use condition variable to sync init when using multiple cores

sodank808:
## Description
This PR fixes a possible timing issue when initializing I2C on a different core.

## Motivation and Context
This change ensures that the I2C init is complete before returning, whereas previously it was up to the caller to ensure enough time has passed. This new automatic behavior should be more robust and less error prone.

NOTE: I also changed the core_id checking logic. In the event a core ID was specified explicitly in a single core usage, it seemed more appropriate to use the direct i2c_driver_install call.

### DIFF
--- a/components/i2c/include/i2c.hpp
+++ b/components/i2c/include/i2c.hpp
@@ -69,7 +69,7 @@ public:
 
     logger_.debug("Initializing I2C with config: {}", config_);
 
-    std::lock_guard<std::mutex> lock(mutex_);
+    std::unique_lock lock(mutex_);
     i2c_config_t i2c_cfg;
     memset(&i2c_cfg, 0, sizeof(i2c_cfg));
     i2c_cfg.sda_io_num = config_.sda_io_num;
@@ -87,7 +87,7 @@ public:
     // Make copy of variables for easy capture
     i2c_port_t i2c_port = config_.port;
     int core_id = config_.isr_core_id;
-    if (core_id < 0) {
+    if (core_id <= 0) {
       err = i2c_driver_install(i2c_port, I2C_MODE_MASTER, 0, 0, 0);
     } else {
       if (core_id > configNUM_CORES - 1) {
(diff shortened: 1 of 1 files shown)


components/i2c/include/i2c.hpp
```diff
@@ -87,23 +87,26 @@ class I2c : public espp::BaseComponent {
     // Make copy of variables for easy capture
     i2c_port_t i2c_port = config_.port;
     int core_id = config_.isr_core_id;
-    if (core_id < 0) {
+    if (core_id <= 0) {
```
finger563, line 90: You don't know that this function will be called on core 0 without checking so I wouldn't do the direct driver install in this case.

components/i2c/include/i2c.hpp
```diff
@@ -87,23 +87,26 @@ class I2c : public espp::BaseComponent {
...
           err = i2c_driver_install(i2c_port, I2C_MODE_MASTER, 0, 0, 0);
+          init_done_cv_.notify_all();
           return true; // stop the task
         },
         .stack_size_bytes = 2 * 1024,
         .priority = 15,
         .core_id = core_id,
       });
       isr_task->start();
+      init_done_cv_.wait(lock);
```
finger563, line 109: Your mutex code is inverted here - you don't need to lock a mutex to notify.

components/i2c/include/i2c.hpp
```diff
@@ -303,6 +306,7 @@ class I2c : public espp::BaseComponent {
   Config config_;
   bool initialized_ = false;
   std::mutex mutex_;
+  std::condition_variable init_done_cv_; ///< Signal for when I2C init is done (when pinned to 2nd core)
```
finger563, line 309: Since this is only used in init, it could exist within the scope of that function, without needing to be in the class and persist for the life of the class.

finger563 (changes requested):
Thanks for fixing the race condition! I don't think I agree with assuming the caller is on core 0 and there are some slight adjustments to the cv code but otherwise it's looking good.

finger563:
I'll probably make a followup PR which adds a reusable function in the task component which does the core check so it doesn't spawn a thread if it's already on the core you want. It'll also include the cv in a blocking variant.

components/i2c/include/i2c.hpp
sodank808: The intention of grabbing the mutex in the task is to synchronize the start to the init task so it doesn't begin until the first task is waiting. Although it may be possible to check the value of CV before calling wait to account for the case where the notify happened first, I feels its more tidy if the mutex ensures the intended execution order.

finger563 (approved):

PR #235, opened 2024-05-15
Add NVS component, wrapping esp-idf nvs_handle.hpp

SamAtBackbone:
Some convenience wrappers for reaching the NVS on ESP. Usable with any namespace and variable. 

## Description

Created an API including calls to:
* Init
* Erase
* Get variables
* Set variables

with the ESP NVS system. 

Created an example app to test and show the API.

Note that Namespace and Key must be <=15 characters because of the underlying system. This would previously cause a panic abort, so we're testing for string length in this API

## Motivation and Context

Convenience wrappers of all the tests etc required to interface to the NVS

## How has this been tested?

Running both my own app that uses this class, and the example app included. 

Example app running and being reset via button:
image

If namespace or key is too long:
image

## Types of changes

- [X] Documentation Update
- [ ] Software change

## Checklist:

- [X] My change requires a change to the documentation.


components/nvs/include/nvs.hpp
finger563, line 1: Can you add doxygen docstrings to the class and member functions in this file? Afterwards, please update `doc/Doxyfile`, `doc/en/index.rst`, and add `doc/en/nvs.rst` accordingly so that this class' documentation is included in the espp documentation.

components/nvs/include/nvs.hpp
```diff
@@ -0,0 +1,116 @@
...
+                logger_.error("Error {%s} committing to NVS!", esp_err_to_name(err));
+                ec = std::make_error_code(std::errc::protocol_error);
+            }
+            break;
+        default:
+            logger_.error("Error {%s} reading!", esp_err_to_name(err));
+            ec = std::make_error_code(std::errc::protocol_error);
+        }
+        return;
+    }
```
finger563, line 106: I might break this up into two functions: `get_var(...)` which doesn't take a default value and does not do a `set_item(...)`/`commit()`, and a `get_or_set(...)` (or some other name that makes sense) which does what you've implemented here. Currently a user of the API has no way to see if there is a key in that namespace without creating it. Similarly, might be good to have a simple `exists(...)` function which returns true or false and uses the `find_key(...)` API.

components/nvs/example/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+
+# add the component directories that we want to use
+set(EXTRA_COMPONENT_DIRS
+  "../../../components/"
+)
+
+set(
+  COMPONENTS
+  "main esptool_py format nvs"
```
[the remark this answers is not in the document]
finger563: Also, looking at the esp-idf it seems it should be depending on `nvs_flash`, not `nvs`?

finger563 (changes requested):

components/nvs/include/nvs.hpp
SamAtBackbone: I didn't generate the docs, it ran into some build issues. Not sure if my doc is generating it or if it's my setup?

components/nvs/include/nvs.hpp
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * @brief Construct a new NVS object.
+     */
+    explicit Nvs()
+      : BaseComponent( "Nvs", espp::Logger::Verbosity::WARN) {}
+
+    /// @brief Initialize the NVS
+    /// @param[out] ec Saves a std::error_code representing success or failure
+    /// @details If the flash init fails because of no free pages or new version, it erases the NVS and re-initializes
+    void init(std::error_code &ec) {
```
finger563, line 31: It seems like either 
1. init shouldn't be a public function, since you're calling it within each of the other functions anyway, or
2. You shouldn't be calling init from within each of the other functions (since presumably they'll return the appropriate error if NVS wasn't initialized and it's just extra code if the user has already called `init()`)

I think my preference is for option (2), but I've not looked into the internals of `nvs_flash_init()` to see how quickly it returns if it was already initialized and how well it's protected against multiple threads calling it at the same time. If it's a quick return and it's protected anyway, then option (1) would be good as well.

finger563: Just checked their code and it seems that
* They do protect against multiple threads calling it simultaneously (locking)
* It is not a trivial return if it is initialized since it goes through their partition manager and it does some lookup and such, so I'd suggest we go with option (2) and remove the init call from each of the other methods.

finger563 (approved):
Awesome, thanks for the convenience class!

PR #252, opened 2024-06-04
fix(nvs): pass ec reference to erase()

guraj:
## Description
Fixed a bug by changing the ec parameter of erase() to pass the error code by reference, allowing it to correctly update the error state within the method.

## Motivation and Context
Prior to this, changes made to the error code by erase() won't affect the original error code outside the method. This fix corrects the signature to reflect the ec convention of the other methods in the Nvs class.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Software change


finger563 (approved):
Thanks, good catch!

PR #257, opened 2024-06-07
feat(nvs): NVSHandle Wrapper

guraj:
## Description
This addition to nvs.hpp is to add functions that allow the user to open an NVS handle, perform multiple get/set operations, and then close the handle. Furthermore, these functions wrap the error handling for better readability for the user.

## Motivation and Context
The existing ESPP wrapper for the NVS component has get/set functions to interact with variables in NVS, but each get/set call inefficiently opens and closes an NVS handle. 

## How has this been tested?
The example app nvs/example/nvs_example.cpp demonstrates the features of this handle, essentially replicating the nvs_rw_value_cxx example app in esp-idf but using ESPP conventions instead.


components/nvs/example/main/nvs_example.cpp
finger563, line 1: @guraj Can you please replace the use of `printf` in this file to use `fmt::print`, since `fmt::print` is the new c++ standard for printing as it provides compile time format syntax and some argument validity checking.

components/nvs/example/main/nvs_example.cpp
```diff
@@ -12,80 +12,145 @@ extern "C" void app_main(void) {
...
+    printf("\n");
+    printf("Opening Non-Volatile Storage (NVS) handle... ");
+    // Handle will automatically close when going out of scope or when it's reset.
+    espp::NVSHandle storage("storage", ec);
+    printf("Done\n");
+    ec.clear();
+    
+    // Read
+    printf("Reading restart counter from NVS ... ");
+    int32_t restart_counter = 0; // value will default to 0, if not set yet in NVS
```
finger563, line 110: This comment isn't true, if I'm reading the NvsHandle implementation correctly

components/nvs/example/main/nvs_example.cpp
```diff
@@ -12,80 +12,145 @@ extern "C" void app_main(void) {
...
+    printf("\n");
+
+    // Restart module
+    for (int i = 10; i >= 0; i--) {
+        printf("Restarting in %d seconds...\n", i);
+        vTaskDelay(1000 / portTICK_PERIOD_MS);
+    }
+    printf("Restarting now.\n");
+    fflush(stdout);
+    esp_restart();
```
finger563, line 151: I believe this code is lifted from esp-idf example, though I think it's better to not have it automatically restart, given that the serial connection provides restart mechanisms. This way you have more than 10 seconds to read the output and validate it's correct :)

components/nvs/include/nvs.hpp
```diff
@@ -70,6 +70,225 @@ const NvsErrCategory theNvsErrCategory{};
 } // namespace
 
 namespace espp {
+class NVSHandle : public BaseComponent {
```
finger563, line 73: Please move this class into `nvs_handle_espp.hpp` in this folder. This will better enable users to only include the code they want, and will enable a future refactor where the `espp::Nvs` class uses `espp::NvsHandle` internally to reduce code duplication.

Additionally, please add a docstring to the class itself as well, similar to the nvs class above, and include a section/snippet pointing to your new example.

This move to a file will require you to update the `doc/Doxyfile` and `doc/en/nvs.rst` to include the new file.

components/nvs/include/nvs.hpp
```diff
@@ -70,6 +70,225 @@ const NvsErrCategory theNvsErrCategory{};
...
+                    ns_name);
+      ec = make_error_code(NvsErrc::Open_NVS_Handle_Failed);
+    }
+  }
+
+  /// @brief Reads a variable from the NVS
+  /// @param[in] key NVS Key of the variable to read
+  /// @param[in] value Variable to read
+  /// @param[out] ec Saves a std::error_code representing success or failure
+  /// @details Read the key/variable pair without committing
```
finger563, line 95: 1. reads don't require / involve committing
2. Probably best to just say read the value of of `key` into `variable`, if `key` exists.

components/nvs/include/nvs.hpp
```diff
@@ -70,6 +70,225 @@ const NvsErrCategory theNvsErrCategory{};
...
+class NVSHandle : public BaseComponent {
+public:
+  /// @brief Construct a new NVSHandle object
+  /// @param[in] ns_name Namespace for NVS
+  /// @param[out] ec Saves a std::error_code representing success or failure
+  /// @details Create an NVSHandle object for the key-value pairs in the ns_name namespace
+  explicit NVSHandle(const char *ns_name, std::error_code &ec)
+      : BaseComponent("NVSHandle", espp::Logger::Verbosity::WARN) {
+    esp_err_t err;
+    ns_name = ns_name;
```
finger563, line 82: You should probably add a check on the ns_name here (though likely the call below will handle that for you).

components/nvs/include/nvs.hpp
```diff
@@ -70,6 +70,225 @@ const NvsErrCategory theNvsErrCategory{};
...
+      ec = make_error_code(NvsErrc::Commit_NVS_Failed);
+    }
+    return;
+  }
+
+protected:
+  std::unique_ptr<nvs::NVSHandle> handle;
+  const char *ns_name;
+
+  void check_lengths(const char *ns_name, const char *key, std::error_code &ec) {
```
finger563, line 272: you should update this function to not take in / check `ns_name`, since that variable doesn't change after construction, but this function is called every time an operation is performed on the handle.

components/nvs/include/nvs.hpp
```diff
@@ -70,6 +70,225 @@ const NvsErrCategory theNvsErrCategory{};
...
+    if (err != ESP_OK) {
+      logger_.error("Error {} committing to NVS!", esp_err_to_name(err));
+      ec = make_error_code(NvsErrc::Commit_NVS_Failed);
+    }
+    return;
+  }
+
+protected:
+  std::unique_ptr<nvs::NVSHandle> handle;
+  const char *ns_name;
```
finger563, line 270: Should change this from being `const char *` to `std::string` so that this class owns the memory associated with it and the handle can be dynamically determined at runtime / at construction instead of constant.

finger563 (changes requested):
Thanks for the update! It's definitely looking better, though I'd prefer the nvs handle code to be in a separate file for clarity. We'll likely have a follow-up PR which expands the functionality some more and which refactors the nvs code to reduce code duplication.

components/nvs/include/nvs.hpp
guraj: I decided to omit the ns_name class variable entirely since the constructor is the only member that uses its value (after modifying the check_lengths() method to not check ns_name). Is there still a good reason to keep it in as a class variable?

components/nvs/example/main/nvs_example.cpp
guraj: That's right, this is an inaccurate comment. Am I understanding correctly that if a key is not found, the method would simply return without modifying restart_counter?

finger563: That's correct

components/nvs/include/nvs.hpp
finger563: Hmm not really. The only reason I can think of is for a caller to later query what namespace the handle was configured for (e.g. if the handle is one of many within another component, but I think such use would be atypical and so I see no need to store the data and add accessor methods for now.

components/nvs/example/main/nvs_example.cpp
```diff
@@ -12,80 +12,136 @@ extern "C" void app_main(void) {
...
+      if (ec.value() == static_cast<int>(NvsErrc::Key_Not_Found)) {
+        fmt::print("The value is not initialized yet!\n");
+      } else if (ec.value() == static_cast<int>(NvsErrc::Read_NVS_Failed)) {
+        fmt::print("Failed to read from NVS: %s\n", ec.message().c_str());
+      } else {
+        fmt::print("An error occurred: %s\n", ec.message().c_str());
+      }
+    } else {
+      fmt::print("Done\n");
+      fmt::print("Restart counter = %" PRIu32 "\n", restart_counter);
```
finger563, line 121: `fmt::` functions don't require older formats like this, and instead use template type deduction, so you can simply use more modern format strings such as

```c++
fmt::print("Restart counter = {}\n", restart_counter);
```

Which would work regardless of whether restart_counter was `uint32_t`, `uint8_t`, `std::string`, `std::vector<uint8_t>`, etc.

components/nvs/example/main/nvs_example.cpp
```diff
@@ -12,80 +12,136 @@ extern "C" void app_main(void) {
...
+    fmt::print("Reading restart counter from NVS ... ");
+    int32_t restart_counter = 0; // value will default to 0, if not set yet in NVS
+    storage.get("restart_counter", restart_counter, ec);
+    if (ec) {
+      if (ec.value() == static_cast<int>(NvsErrc::Key_Not_Found)) {
+        fmt::print("The value is not initialized yet!\n");
+      } else if (ec.value() == static_cast<int>(NvsErrc::Read_NVS_Failed)) {
+        fmt::print("Failed to read from NVS: %s\n", ec.message().c_str());
+      } else {
+        fmt::print("An error occurred: %s\n", ec.message().c_str());
```
finger563, line 117: Similar comment as elsewhere - you should replace `%s` with `{}` in your format string.

components/nvs/include/nvs.hpp
finger563, line 1: Since these are whitespace only changes, please revert them.

components/nvs/include/nvs_handle_espp.hpp
finger563, line 1: Your constructor can return without initializing your `handle` variable, but the other APIs don't check the `handle` variable before dereferencing it, which can lead to runtime crashes should the caller improperly ignore the error code from the constructor.

Please add checks to the other methods to ensure that `handle` is valid before continuing with the function. This would also likely necessitate another error code added for invalid or uninitialized handle.

components/nvs/include/nvs.hpp
finger563: Related - given that there _are_ whitespace changes, can you please ensure that you have set up the `pre-commit` hook for this repo? I apologize as I've not added those instructions to the README or to a CONTRIBUTING file, so here they are (copied from https://pre-commit.com):

```sh
# ensure clang-format is installed on your system; for macos:
brew install clang-format

# for your first time using pre-commit:
# use this:
pip install pre-commit
# or this:
pipx install pre-commit

# first time setting up pre-commit for this repo, from the root of the repo:
pre-commit install

# Since you've made commits / changes before installing the pre-commit hook, you should also run this once:
pre-commit run --all-files

# From now on, any time you make commits to espp, pre-commit will run and will ensure that clang format is applied. If any violations are found, the files will be changed to match the format rules and you will need to re-commit them.
```

finger563 (changes requested):
Thanks for making those changes! I've found another issue related to the handle possibly dereferencing a nullptr, so please fix that. There's also some style updates that should be easy to apply once you set up clang-format and pre-commit (I've provided the instructions in a comment)

components/nvs/example/main/nvs_example.cpp
finger563, line 1: Also, it appears you've not added an include for the `nvs_handle_espp.hpp` file to this example, so the example fails to build.

finger563 (approved):

Issue #265, opened 2024-07-01
Migrating LVGL from v8 to v9


abrondijk:
### Answers checklist.

- [X] I have read the documentation and the issue is not addressed there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I'm working on migrating LVGL from v8 to v9 and I've run into some issues in regards to your `espp::Display` class.

LVGL v9 no longer uses macro magic to populate the `lv_color_t` struct depending on the chosen color depth. Instead, it has different types for the different color depths, and `lv_color_t` represents RGB888.
This means that `lv_color_t` can no longer be the type of the framebuffers if the display in question uses the RGB565 format for example.

Now this is all fine and can be implemented in the different display driver classes, but currently, the `espp::Display` class uses `lv_color_t` as well.

The `vram` pointers can be void pointers (also what LVGL uses internally), but the `vram_size_bytes()` function uses `sizeof(lv_color_t)`.
Should the `AllocatingConfig` and the `NonAllocatingConfig` structs also have a member that holds the buffer size in bytes, like the `pixel_buffer_size`?
Or, perhaps a member that holds the amount of bytes per pixel?

Additionally, I think that the `uint16_t *vram0()` and `uint16_t *vram1()` functions should instead also be void pointers to support all color depths.

Alternatively, the `espp::Display` class could be templated to indicate the pixel type, but that would be a rather large refactor.

What are your thoughts?

finger563:
@abrondijk Thanks for looking into this - I've wanted to look into moving to v9, but haven't had the time to do so yet 😅 

Regarding your idea - I think I prefer the template approach because I think that will keep the API cleaner and more strongly typed, while still allowing for support of both v8 and v9 should someone want to stay on v8.

abrondijk:
@finger563 Ok, I'll start working on that!

As for being able to use both versions, the way to tie the framebuffers and the flush callbacks are a bit different. If you'd like to support both, I can probably work around it with macros.

finger563:
@abrondijk Personally, I'd prefer to simply migrate to v9 (since that's what would be bundled with espp anyway) if that's the case.

PR #282, opened 2024-07-07
(#265) LVGL v9 & `espp::Display` 

abrondijk:
## Description

Updates the LVGL submodule to their latest release tag (v9.1), and implements #265.

## Motivation and Context

Easier to maintain when up to date, and the new major release includes a lot more widgets.

## How has this been tested?

I've tested these changes by running the example with the included Smartknob-HA configuration on the actual Smartknob-HA, as well as by using the new `Display` class in the Smartknob-HA firmware.

However, I'm unable to test all the other hardware configurations in the example, as I don't own any of those devkits.

The same goes for the T-Deck, T-Dongle and the ESP-BOX.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update
- [X] Software change

## Checklist:


finger563:
Awesome, thanks @abrondijk! I'll test these changes on some of the other examples here shortly!

abrondijk:
@finger563 I forgot to mention, I've also been unable to test the `input_driver` component entirely, as I have no touch capable device, or device with buttons. The changes seem minor, but you never know.

components/display_drivers/example/main/gui.hpp
```diff
@@ -43,15 +41,15 @@ class Gui {
 protected:
   void init_ui() {
     // Create a container with COLUMN flex direction
-    column_ = lv_obj_create(lv_scr_act());
-    lv_obj_set_size(column_, display_->width(), display_->height());
+    column_ = lv_obj_create(lv_screen_active());
+    lv_obj_set_size(column_, LV_HOR_RES, LV_VER_RES);
```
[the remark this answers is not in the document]
finger563: Perhaps if we don't want to bother with storing a ptr to the display, we could add the width/height to the config, and pass it through to the init_ui function?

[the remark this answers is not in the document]
finger563: Otherwise we could just use the apis to get the width/height of the `lv_screen_active`

components/display/include/display.hpp
finger563, line 1: Now that this is a template class, it might be beneficial to move the `Rotation`, `Signal` enums to a `espp::display` namespace or move them into the `espp` namespace as `espp::DisplayRotation` and `espp::DisplaySignal`. It requires a little more breaking changes, so I'm not convinced we _have_ to do it but it would make accessing them easier since they don't depend on the template type.

components/display/include/display.hpp
```diff
@@ -54,7 +48,8 @@ class Display : public BaseComponent {
     size_t width;             /**< Width of th display, in pixels. */
     size_t height;            /**< Height of the display, in pixels. */
     size_t pixel_buffer_size; /**< Size of the display buffer in pixels. */
-    flush_fn flush_callback;  /**< Function provided to LVGL for it to flush data to the display. */
+    lv_display_flush_cb_t
```
finger563, line 51: Can we do a `using flush_cb_t = lv_display_flush_cb_t` in this class so that if this changes in the future, there are fewer changes we need to make :)

finger563 (commented):
Thanks for the PR! It looks really good, though I think there are a few more updates potentially needed to the examples and I have some suggestions for discussion. Now that I've looked through it, I'll pull and try to test on esp-box and t-deck to test the examples and the input drivers.

finger563:
@abrondijk Is there a change to the sdkconfig for the themeing? I'm noticing color change when testing on your branch. In good news, I did verify that the `esp-box/example` works (meaning touch input works just fine :) )

finger563:
The `t-dongle-s3/example` does not work - it builds and flashes, but the screen is not configured / operating correctly.

components/display/include/display.hpp
```diff
@@ -269,23 +265,20 @@ class Display : public BaseComponent {
...
+
     // Configure the LVGL display buffer with our pixel buffers
-    lv_disp_draw_buf_init(&disp_buffer_, vram_0_, vram_1_, display_buffer_px_size_);
+    lv_display_set_buffers(display_, vram_0_, vram_1_, vram_size_bytes(),
+                           LV_DISPLAY_RENDER_MODE_PARTIAL);
 
-    lv_disp_drv_init(&disp_driver_);
-    disp_driver_.draw_buf = &disp_buffer_;
-    disp_driver_.flush_cb = flush_callback;
-    disp_driver_.sw_rotate = (uint8_t)sw_rotation_enabled;
```
finger563, line 282: Is this not a configuration that can be used anymore? I'm noticing that some of the examples aren't rotated correctly anymore.

finger563:
> The `t-dongle-s3/example` does not work - it builds and flashes, but the screen is not configured / operating correctly.

I think this may be the same issue as the t-devk issue and is related to rotation support

abrondijk:
@finger563 Thanks for taking the time!
I agree on all your points, I'll apply them in a little bit 

Strange that rotation doesn't seem to work correctly, I'll look into it

components/t-deck/include/t-deck.hpp
```diff
@@ -228,7 +228,7 @@ class TDeck : public BaseComponent {
   /// \note This method queues the data to be written to the LCD, only blocking
   ///      if there is an ongoing SPI transaction
   void write_lcd_frame(const uint16_t x, const uint16_t y, const uint16_t width,
-                       const uint16_t height, const uint8_t *data);
+                       const uint16_t height, uint8_t *data);
```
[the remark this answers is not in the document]
abrondijk: This function passes `data` along to `DisplayDriver::fill`, which is also called from the flush callback function. `lv_display_flush_cb_t` expects a nonconst, so I think this indeed can't be a const anymore.

finger563:
> > @abrondijk Is there a change to the sdkconfig for the themeing? I'm noticing color change when testing on your branch.
> 
> Is there any chance you could upload a photo of your GUI demo of what it is supposed to look like?
> I've actually been struggling with theming since v8 already, so I'm not 100% sure what the demo is even supposed to look like color wise...
> 
> 

@abrondijk the readmes for those examples have images / videos that show what it should look like. The colors / themes for all those examples are the same so the video in t-dongle-s3 is representative of them all

finger563:
Also, @abrondijk I meant to ask - have you tested out any of the performance improvements that might come to LVGL v9 on ESP32?

Reference: https://github.com/lvgl/lvgl/issues/4926#issuecomment-1905860994

finger563:
Also, the migration section of the [changelog](https://github.com/lvgl/lvgl/blob/release/v9.0/docs/CHANGELOG.rst) says that we should update the copied lv-conf template (which is the lv-conf.h in display component)

abrondijk:
@finger563 Oh that sounds really promising, I had not seen that comment or those changes.

I did see that you can now enable OS support, but I've yet to have a successful build with freertos enabled..

I think this might warrant a subsequent PR though, the move from v8 to v9 is a bit more of a pain than I initially thought.
I'll work on those 2 features when this is finished though!

Might take a little bit however, I'm a bit low on time this week.

abrondijk:
@finger563 It seems as though software rotation has been stripped from lvgl v9, and should now be implemented by the user in the flush callback, using `lv_draw_sw_rotate`

This begs the question whether we would even want that to be implemented at all, as surely hardware rotation is more efficient.

All that `lv_display_set_rotation` does right now, is swap the internal horizontal and vertical resolutions accordingly, and create an event.

What I would suggest is that the `espp::Display` class, registers an lvgl event handler.
The handler would then call `setRotation` which is passed in the constructor (like `flush_cb`), that writes the necessary control bits over spi.
`setRotation` would be implemented in the different display drivers. Its logic would be basically the same as the logic around the `mirror_x` and `mirror_y` in `initialize` of the respective display drivers.

What do you think of this approach?

> Also, the migration section of the [changelog](https://github.com/lvgl/lvgl/blob/release/v9.0/docs/CHANGELOG.rst) says that we should update the copied lv-conf template (which is the lv-conf.h in display component)

I've updated the lv_conf as you suggested, but I'm not entirely sure what the purpose is of the file there, as I don't think it's actually used anywhere.
Through the different Kconfig options, it does not seem to be possible to set the lv_conf path to this file, and besides that, it's not ideal having to make changes to a sourcefile of a library to change settings.

finger563:
> @finger563 It seems as though software rotation has been stripped from lvgl v9, and should now be implemented by the user in the flush callback, using `lv_draw_sw_rotate`
> 
> This begs the question whether we would even want that to be implemented at all, as surely hardware rotation is more efficient.
> 
> All that `lv_display_set_rotation` does right now, is swap the internal horizontal and vertical resolutions accordingly, and create an event.
> 
> What I would suggest is that the `espp::Display` class, registers an lvgl event handler.
> The handler would then call `setRotation` which is passed in the constructor (like `flush_cb`), that writes the necessary control bits over spi.
> `setRotation` would be implemented in the different display drivers. Its logic would be basically the same as the logic around the `mirror_x` and `mirror_y` in `initialize` of the respective display drivers.
> 
> What do you think of this approach?
> 
> > Also, the migration section of the [changelog](https://github.com/lvgl/lvgl/blob/release/v9.0/docs/CHANGELOG.rst) says that we should update the copied lv-conf template (which is the lv-conf.h in display component)
> 
> I've updated the lv_conf as you suggested, but I'm not entirely sure what the purpose is of the file there, as I don't think it's actually used anywhere.
> Through the different Kconfig options, it does not seem to be possible to set the lv_conf path to this file, and besides that, it's not ideal having to make changes to a sourcefile of a library to change settings.

@abrondijk thanks for looking into it and evaluating it - I agree that hardware rotation is the preferred solution, so I'll look at making sure our existing drivers support it and I agree that the additional callback should be provided and architected in that way. 

Also agreed that lv_conf shouldn't be modified, and if it's not used by the build or lvgl configuration anymore we should remove it for clarity, though I thought the cmakelists did enable it to be included by lvgl. Would love to remove it if possible.

components/esp-box/example/main/esp_box_example.cpp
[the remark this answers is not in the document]
abrondijk: Technically no, all the previous API functions are typedeffed to the new ones, so they will still work.

I've updated them anyway now though!

abrondijk:
@finger563 I have implemented the eventhandler for hardware rotation.
So far I have only included it in the GC9A01 ([here](https://github.com/smartknob-ha/espp/blob/fc83129342441e7d4873d21ceb783628b93c1706/components/display_drivers/include/gc9a01.hpp#L190-L212)) as that is the only display controller that I own, but I don't imagine the implementation being any different for the other controllers.

I have also moved the enums from inside `espp::Display`, to outside.

I made the function pointer null by default, so it can just not be passed to the display class.

finger563:
> @finger563 I have implemented the eventhandler for hardware rotation.
> So far I have only included it in the GC9A01 ([here](https://github.com/smartknob-ha/espp/blob/fc83129342441e7d4873d21ceb783628b93c1706/components/display_drivers/include/gc9a01.hpp#L190-L212)) as that is the only display controller that I own, but I don't imagine the implementation being any different for the other controllers.
> 
> I have also moved the enums from inside `espp::Display`, to outside.
> 
> I made the function pointer null by default, so it can just not be passed to the display class. 

@abrondijk awesome, I'll test it out this evening on the displays I have and provide updates here

finger563:
Hey @abrondijk,

I've tested the examples on the hardware and there are a few compilation issues and other changes (now that we're not using software rotation). I've attached the patch file to this comment so that you can commit it. The theme is still weird (the circles / edges are supposed to be grey instead of green), so I'm not sure if that's a theme issue or a color representation issue (since I know that our LVGL had some color swapping configured).

Also, can you please merge in `espp/main` since it's got some changes to the demos that I think are worth including (and it adds a `matouch-rotary-display` component+example that should be updated.

image


[fix.patch](https://github.com/user-attachments/files/16229661/fix.patch)

finger563:
It's also hard to see from the image, but the text is harder to read. Might be a theme or color issue as well though

finger563:
I think it has to do with the color swap sdk config that we had set in the defaults for the examples, my guess is that we either don't need it anymore or it's no longer working.

finger563:
Looks like we need a new parameter for swapping the color bytes, since:

> `LV_COLOR_16_SWAP` is removed and `lv_draw_sw_rgb565_swap()` can be called manually in the flush_cb if needed to swap the in-place.

According to the v9 changelog

finger563:
Confirmed:
image

This also means that in the examples' `sdkconfig.defaults`, we can remove the `Color Settings` section values, since they are not used anymore.

E.g.
CleanShot 2024-07-15 at 08 28 29

finger563:
@abrondijk if you'd like, I can take over this PR if you don't have time to finish it - i can fork your fork and submit an updated pr here if you like. I appreciate the work you've done on this and I'm not trying to rush you, just trying to be helpful :)

abrondijk:
> @abrondijk if you'd like, I can take over this PR if you don't have time to finish it - i can fork your fork and submit an updated pr here if you like. I appreciate the work you've done on this and I'm not trying to rush you, just trying to be helpful :)

I appreciate that! 
Motivation has just been on the low side, but I was planning on getting to it again on the weekend

abrondijk:
@finger563 Indeed adding `lv_draw_sw_rgb565_swap` seems to solve the theming/colour issues.
I've gone through some popular display controllers to see if there is some register that would allow solving this in hardware, but I can't find anything like this.

So unfortunately, I can't think of a better way to handle this than by creating a custom spi function that replaces [this](https://github.com/smartknob-ha/espp/blob/e1edf6aa5c9ebde5b1d5ba82c197fb0da32a474f/components/display_drivers/example/main/display_drivers_example.cpp#L163) call.

finger563:
@abrondijk given that it's display, specific, I think I'd prefer to add it as a call within the display driver's `fill` command, since it's mainly an lvgl issue given how they represent colors. Don't want to have to call that for all spi transactions.

abrondijk:
@finger563 I agree!
Do you know whether all currently supported controllers need this call? At least the GC901A and the ST7789

finger563:
> @finger563 I agree!
> Do you know whether all currently supported controllers need this call? At least the GC901A and the ST7789

@abrondijk yes they all do in my testing

abrondijk:
@finger563 I have added the rotation function to the other 2 drivers as well (exact same function as the register is the same), and updated the examples etc.

finger563:
@abrondijk it looks like the update overwrote some of the changes to the hal components like wrover kit, t-display, etc. If you fix that it should be good! I'm not home right now, but when I get a chance later today I'll test the updates on all the displays I've got and let you know how it goes!

components/display/include/display.hpp
```diff
@@ -256,36 +264,48 @@ class Display : public BaseComponent {
...
 
 protected:
+  /**
+   * @brief LVGL event handler.
+   */
+  static void event_cb(lv_event_t *event) {
+    if(event->code == LV_EVENT_RESOLUTION_CHANGED) {
+      auto rotation = lv_display_get_rotation(lv_display_get_default());
+      auto rotation_callback = reinterpret_cast<rotation_fn>(lv_event_get_user_data(event));
+      rotation_callback(static_cast<DisplayRotation>(rotation));
```
finger563, line 277: I know the `event_cb` is only registered in `init` if the callback function is not nullptr, but that may change in the future (e.g. if we support adding a rotation callback setter later), so best to go ahead and check that it's not nullptr here before calling it :)

finger563 (approved):
Awesome, thanks for getting this together! The code looks good, and I'll test it on my displays soon :)

finger563:
I've checked these examples, and they all work, though there are some modifications needed. I've attached a patch file that should make it easy for you to apply :)

- [x] `esp-box/example` ✅ 
- [x] `matouch-rotary-display/example` ✅ 
- [x] `t-deck/example` ✅ 
- [x] `t-dongle-s3/example` ✅ 
- [x] `wrover-kit/example` ✅ 

[final_fix.patch](https://github.com/user-attachments/files/16405788/final_fix.patch)

components/display/include/display.hpp
abrondijk: You're right, that's indeed better. Fixed!

abrondijk:
@finger563 Thank you for taking the time to test and help on this one!

I've applied your patch and resolved your comment

PR #299, opened 2024-08-07
Only do thread join when we know thread is running

sodank808:
## Description
Check the started_ flag before trying to stop/join the thread.

## Motivation and Context
Redundant calls to task stop seem to be prone to exceptions down in std::thread::join. This doesn't quite make sense because of the joinable call ahead of time

### DIFF
--- a/components/task/include/task.hpp
+++ b/components/task/include/task.hpp
@@ -272,10 +272,12 @@ public:
    */
   bool stop() {
     logger_.debug("Stopping task");
-    started_ = false;
-    cv_.notify_all();
-    if (thread_.joinable()) {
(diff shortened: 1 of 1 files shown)


finger563:
Hmm this seems ok - best guess right now for why you might have seen issues is if the stop is called simultaneously from multiple threads since then maybe the joinable call will show true for both threads before the thread is actually joined and exited. Might be good to add a test case for that in the example - I'll put that on a to-do list for when I'm back. I do remember having this code in there and removing it so I'll need to check my notes and tests as well to make sure the thread is actually stopped under all scenarios.

finger563 (approved):

PR #372, opened 2025-02-16
feat(bldc_haptics): add set_position function

SquaredPotato:
## Description

Add a function set_position for use with the bldc_haptics class so we can change the current "haptic position" to any valid position within the current DetentConfig.

## Motivation and Context

Say you have a pre-existing value that you would like to change using, for example, the `COARSE_VALUES_STRONG_DETENTS` config, and the currently stored value is 28.
With the current logic the `current_position_` will always be set on the position of the motor shaft, which might not be equal to our desired starting point of 28.

## How has this been tested?

Calling this function when initializing a config so that it always starts "midway" in the DETENT config, regardless of the position I turn the shaft to before rebooting. (https://github.com/smartknob-ha/firmware/blob/ef9ff84e42428ef23f151100f5b5302cec8a44cd/src/main.cpp#L82)

Tested on a Smartknob-HA 

## Types of changes

- [ ] Documentation Update
- [ ] Software change

## Checklist:

### Software

### DIFF
--- a/components/bldc_haptics/include/bldc_haptics.hpp
+++ b/components/bldc_haptics/include/bldc_haptics.hpp
@@ -186,24 +186,25 @@ public:
   /// @return Current position of the haptic motor
   float get_position() const { return current_position_; }
 
+  /// @brief Sets the current position of the haptic motor to
+  ///        within the bounds of the current detent config
+  /// @param position Position to set the haptics to
+  void set_position(const int position) {
+    current_position_ = std::clamp<int>(position, detent_config_.min_position, detent_config_.max_position);
+  }
+
   /// @brief Configure the detents for the haptic motor
+  /// @param config Detent config
   void update_detent_config(const detail::DetentConfig &config) {
     std::unique_lock<std::mutex> lk(detent_mutex_);
     // update the detent center
     current_detent_center_ = motor_.get().get_shaft_angle();
 
-    if (current_position_ < config.min_position) {
-      // if the current position is less than the min position, set the current
-      // position to the min position
-      current_position_ = config.min_position;
-    } else if (current_position_ > config.max_position) {
-      // if the current position is greater than the max position, set the
-      // current position to the max position
-      current_position_ = config.max_position;
-    }
-
     // update the detent config
     detent_config_ = config;
+
+    // set position after setting the config
+    set_position(current_position_);
 
     // Update derivative factor of torque controller based on detent width. If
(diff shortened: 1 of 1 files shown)


finger563 (approved):
Thanks! Looks good to me!  Can you please add @param doxy comments for the new function (and the missing one for the old function 😅)

components/bldc_haptics/include/bldc_haptics.hpp
```diff
@@ -186,25 +186,34 @@ template <MotorConcept M> class BldcHaptics : public BaseComponent {
...
-  /// @brief Configure the detents for the haptic motor
-  void update_detent_config(const detail::DetentConfig &config) {
-    std::unique_lock<std::mutex> lk(detent_mutex_);
-    // update the detent center
-    current_detent_center_ = motor_.get().get_shaft_angle();
-
-    if (current_position_ < config.min_position) {
+  /// @brief Sets the current position of the haptic motor to
+  ///        within the bounds of the current detent config
+  void set_position(const int position) {
```
finger563, line 191: Also, might be cleaner code to use std::clamp for this instead of conditional code

finger563 (approved):

components/bldc_haptics/include/bldc_haptics.hpp
SquaredPotato: For sure

components/bldc_haptics/include/bldc_haptics.hpp
```diff
@@ -186,25 +186,26 @@ template <MotorConcept M> class BldcHaptics : public BaseComponent {
   /// @return Current position of the haptic motor
   float get_position() const { return current_position_; }
 
+  /// @brief Sets the current position of the haptic motor to
+  ///        within the bounds of the current detent config
+  /// @param position Position to set the haptics to
+  void set_position(const int position) {
+    current_position_ = std::clamp<int>(detent_config_.min_position, detent_config_.max_position, position);
```
finger563, line 193: Your clamp is out of order - it's value, low, high

finger563 (changes requested):
Thanks, though your order for the clamp function is wrong - once you fix that we should be good to merge

components/bldc_haptics/include/bldc_haptics.hpp
SquaredPotato: Man I don't understand how I got this wrong, sharp

finger563 (approved):

PR #383, opened 2025-02-23
Display refactor and SH8601 support

abrondijk:
## Description

The main purpose of this PR is to add support for the [SH8601](https://dl.espressif.com/AE/esp-iot-solution/SH8601A0_DataSheet_Preliminary_V0.0_UCS__191107_1_.pdf) OLED display driver.

As this is an inherently different display driver, it requires a few changes to the display and display_driver components:

 - Refactored `LcdInitCmd` struct to have configurable delays per command.
 - Moved `set_brightness` & `get_brightness` functions to the display driver itself.
   This is because OLED displays don't have a backlight of course, and thus its brightness is controlled through the SPI interface rather than a PWM signal.
 - Make implementing the `send_lines` function a requirement for the following:
 - Change `lcd_write` function to a dedicated `write_command` function.
   For the currently supported display drivers, sending commands is rather trivial. But this OLED driver (and from what I've found so far, a lot of QSPI drivers), handles sending commands quite a bit differently compared to writing to gram.
 - Have the DC pin be optional
   This OLED driver does not have a DC signal exposed

## Motivation and Context

For [Smartknob-HA](https://github.com/smartknob-ha/), we are exploring the possibility of swapping the display to this OLED one.

## How has this been tested?

This has been tested on the [Lilygo T-Encoder Pro](https://lilygo.cc/products/t-encoder-plus).

The changes above, and the actual QSPI functions, can be found [here](https://github.com/smartknob-ha/firmware/blob/912bf0ba61c1867520573e2fbc19f395ff3ac5ce/components/display_driver/source/DisplayDriver.cpp), mainly the `writeCommand`, `sendLines` and `initialize` functions.

If you agree to the changes I've made to the components, I will update the driver example, and also apply the above changes to the already existing drivers and components using them.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update
- [X] Software change

## Checklist:

- [X] My change requires a change to the documentation.

### Software

### Hardware


abrondijk:
@finger563 Let me know what you think :)

finger563:
@abrondijk this sounds like a good change - please go ahead and update the other display drivers accordingly and once you lmk it's ready I'll test on my systems. This design pattern should work well with my recent change removing gamma configuration from the driver itself since that may be display / board dependent. Def like the idea of officially adding OLED support to espp 🚀

finger563:
@abrondijk is there a way the LED could still be part of the display, but just be optional? such that we don't use the LED / initialize at all if the provided GPIO_NUM is -1? That way it's still part of the display rather than the display driver?

abrondijk:
@finger563 It is possible, but that would mean another 2 functions to pass to the display constructor, for displays that don't have a PWM controlled backlight.

As it is right now, I'm not entirely happy anyway, with the [LED instance having to be a pointer](https://github.com/esp-cpp/espp/blob/52ac665d43eecebd8a4e1cd5c750ed1daa41b5f2/components/display_drivers/include/gc9a01.hpp#L114-L117) (because of the lack of constructor).
I'll try to find a better solution.

abrondijk:
@finger563 In my most recent commit, I've moved the brightness back to the display component, and added 2 functions to the config struct.
Does this seem alright to you?

abrondijk:
On another note, I'm not entirely sure whats causing the static analysis to fail. Do you have any idea?

finger563:
> On another note, I'm not entirely sure whats causing the static analysis to fail. Do you have any idea?

I think I've probably got it misconfigured - I think it's complaining because it can't find the file in the upstream repo since this PR is from a fork. It's worked on other PRs from forks before so I think it just has to do with this PR adding a new file.

I thought I'd configured it properly to analyze fork PRs but I'll take another look.

components/display/include/display.hpp
```diff
@@ -62,10 +74,16 @@ template <typename Pixel> class Display : public BaseComponent {
     size_t pixel_buffer_size; /**< Size of the display buffer in pixels. */
     flush_fn flush_callback;  /**< Function provided to LVGL for it to flush data to the display. */
     rotation_fn rotation_callback{
-        nullptr};             /**< Function used to configure display with new rotation setting. */
-    gpio_num_t backlight_pin; /**< GPIO pin for the backlight. */
+        nullptr}; /**< Function used to configure display with new rotation setting. */
+    gpio_num_t backlight_pin{GPIO_NUM_NC}; /**< GPIO pin for the backlight. */
```
finger563, line 78: Please indicate in the comment that this is an optional parameter

components/display/include/display.hpp
```diff
@@ -62,10 +74,16 @@ template <typename Pixel> class Display : public BaseComponent {
...
     bool backlight_on_value{
         true}; /**< Value to write to the backlight pin to turn the backlight on. */
+    set_brightness_fn set_brightness_callback{
+        nullptr}; /**< Callback for setting the display brightness if it can't be controlled by
+                     simple pwm. */
+    get_brightness_fn get_brightness_callback{
+        nullptr}; /**< Callback for getting the display brightness if it can't be controlled by
+                    simple pwm. */
```
finger563, line 86: So one way i think we can handle this is to actually break apart the `Config` struct into two different config structs. One that has the gpio num (no default, so it must be provided), and another which has these functions and no gpio. We can name them differently so it's clear what their purpose is and it makes the caller code less possible to misconfigure the display.

finger563 (commented):
I agree with your comments on cluttering the display config - so I think what we can do is similar to what we did for tasks and other components: multiple configs. This will make the caller code clearer since we can have a config specifically for LED PWM, and a config specifically for more general control with function callbacks.

Hope that makes sense and keeps the code easy to use and (hopefully) harder to misuse.

components/display/include/display.hpp
abrondijk: Sounds like a plan. Do you want to include that in this PR?

finger563: Yes please, if that's not too much work

abrondijk: No that's ok. Do you have any preference for what the new Config objects look like?

finger563: Id recommend simply mirroring the fields in the config we have, but removing the ones that don't make sense for the new display types and adding the new fields we want. 

Given that these are replacing the led pwm display with generic functions id probably call the new one GenericConfig? Then we just need a new constrictor which takes that as the config type

finger563: We could split the configs in two so that it's a little cleaner - we have the hardware configs which would be for LCD and OLED and have the backlight and such, then we have the memory configs which would be the allocating or non allocating configs. Then it's cleaner in the future if we added other display types such as eink or went crazy and did a laser or something

abrondijk: What do you think? Is this to your liking?

components/display/include/display.hpp
```diff
@@ -356,8 +443,10 @@ template <typename Pixel> class Display : public BaseComponent {
   Pixel *vram_1_{nullptr};
   bool created_vram_{false};
   std::vector<Led::ChannelConfig> led_channel_configs_;
-  Led backlight_;
+  std::unique_ptr<Led> backlight_;
   std::chrono::duration<float> update_period_;
   lv_display_t *display_;
+  set_brightness_fn set_brightness_;
+  get_brightness_fn get_brightness_;
```
finger563, line 450: Please init these to nullptr explicitly here

finger563 (commented):
Thanks, this looks good! I've done a quick read through the API changes and they look good, so I'll try to do a full review here a little bit later today!

components/display/include/display.hpp
```diff
@@ -53,125 +54,182 @@ template <typename Pixel> class Display : public BaseComponent {
...
+  struct DynamicMemoryConfig {
+    size_t pixel_buffer_size; /**< Size of the display buffer in pixels. */
     bool double_buffered{
         true}; /**< Whether to use double buffered rendering (two display buffers) or not. */
     uint32_t allocation_flags{
         MALLOC_CAP_8BIT |
         MALLOC_CAP_DMA}; /**< For configuring how the display buffer is allocated*/
-    DisplayRotation rotation{
-        DisplayRotation::LANDSCAPE}; /**< Default / Initial rotation of the display. */
-    bool software_rotation_enabled{
```
finger563, line 82: i don't see this in any of the new configs, do we not want to support this anymore?

components/display/include/display.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include "base_component.hpp"
 #include "led.hpp"
+#include "logger.hpp"
```
finger563, line 13: technicall you don't need this since the include comes from `base_component.hpp`

components/display/include/display.hpp
```diff
@@ -53,125 +54,182 @@ template <typename Pixel> class Display : public BaseComponent {
...
+                   const StaticMemoryConfig &mem_conf,
+                   const Logger::Verbosity log_level = Logger::Verbosity::WARN)
+      : BaseComponent("Display", log_level)
+      , width_(lvgl_conf.width)
+      , height_(lvgl_conf.height)
+      , display_buffer_px_size_(mem_conf.pixel_buffer_size)
+      , vram_0_(mem_conf.vram0)
+      , vram_1_(mem_conf.vram1)
+      , update_period_(lvgl_conf.update_period) {
+    init(lvgl_conf.flush_callback, lvgl_conf.rotation_callback, lvgl_conf.rotation,
```
finger563, line 149: might make it cleaner to separate out this `init` function into `init_gfx` and `init_backlight`, then in the OLED / non backlit case we simply don't call `init_backlight`

components/display/include/display.hpp
abrondijk: `software_rotation_enabled` was not being used anywhere in the code, so I figured I'd remove it from the config struct

components/display/include/display.hpp
```diff
@@ -285,20 +353,39 @@ template <typename Pixel> class Display : public BaseComponent {
...
    * @param task_config Configuration for the task that runs the lvgl tick
+   * @param backlight_pin GPIO pin for the backlight. -1 if no backlight.
+   * @param backlight_on_value Value to write to the backlight pin to turn the backlight on.
    */
-  void init(flush_fn flush_callback, rotation_fn rotation_callback, DisplayRotation rotation,
-            const Task::BaseConfig &task_config) {
+  void init(const flush_fn flush_callback, const rotation_fn rotation_callback,
+            DisplayRotation rotation, const Task::BaseConfig &task_config,
+            const gpio_num_t backlight_pin = GPIO_NUM_NC, const bool backlight_on_value = true) {
+    if (backlight_pin != -1) {
```
finger563, line 365: I know they're the same, but please make this a check against `GPIO_NUM_NC` instead of `-1`

components/display/include/display.hpp
finger563, line 1: I've noticed in the constructors you've removed the call to `set_brightness(1.0);` I'd like to leave that so as not to break existing functionality as much which expected the display to initialize to have the backlight / brightness full (if backlight config information was provided). Can you please add that call after the call to `init` in each of the constructors?

components/display/include/display.hpp
finger563: Follow-up: just looked through the code and it seems we don't use it anymore anyway, so thanks for removing it :)

components/display/include/display.hpp
```diff
@@ -285,20 +353,39 @@ template <typename Pixel> class Display : public BaseComponent {
...
+                                      .timer = LEDC_TIMER_0,
+                                      .output_invert = !backlight_on_value});
+
+      backlight_ = std::make_unique<Led>((Led::Config{.timer = LEDC_TIMER_0,
+                                                      .frequency_hz = 5000,
+                                                      .channels = led_channel_configs_,
+                                                      .duty_resolution = LEDC_TIMER_10_BIT}));
+    } else {
+      if (set_brightness_ == nullptr || get_brightness_ == nullptr) {
+        logger_.error("No backlight pin provided and no brightness control callbacks provided!");
```
finger563, line 377: I think we should consider this a warning, rather than an error.

components/display_drivers/include/display_drivers.hpp
```diff
@@ -8,11 +8,14 @@ namespace espp {
...
  * @brief Low-level callback to write bytes to the display controller.
- * @param data Pointer to array of bytes to write.
- * @param length Number of bytes to write.
+ * @param command to write
+ * @param data Pointer to array of command parameters to write.
+ * @param length Number of command parameters to write.
  * @param user_data User data associated with this transfer, used for flags.
  */
-typedef std::function<void(const uint8_t *data, size_t length, uint32_t user_data)> write_fn;
+typedef std::function<void(uint8_t command, const uint8_t *parameters, size_t length,
```
finger563, line 16: Given that down below you have a template where you could specify a command that is `std::uint16_t`, I think it makes sense to update this typedef to be templated as well. 

If we're wanting to support this, it might make the most sense to make the display_drivers namespace a template class, which the display drivers are then subclass + specializations of. 

This would allow something like:

```c++
class St7789 : public espp::DisplayDriver<std::uint8_t> { ... }
```

Which makes it cleaner and easier to define a display driver that may take a `uint8_t` command or one which takes a `uint16_t` command.

finger563: Note that if we go this route, to make application code simpler, i'd add some `using`s to the top of the subclasses, such as:

```c++
class St7789 : public espp::DisplayDriver<std::uint8_t> {
public:
  using write_command_fn = espp::DisplayDriver<std::uint8_t>::write_command_fn;
  using LcdInitCmd = espp::DisplayDriver<std::uint8_t>::LcdInitCmd;

...
```

finger563: This is a similar pattern to what we do when we subclass `espp::Interrupt` into `espp::Button`.

components/display_drivers/include/display_drivers.hpp
```diff
@@ -32,13 +35,11 @@ typedef std::function<void(int sx, int sy, int ex, int ey, const uint8_t *color_
...
-  send_lines_fn lcd_send_lines{
-      nullptr}; /**< Function which the display driver uses to send bulk (color) data (non-blocking)
-                   to be written to the display. If not provided, it will default to using the
-                   provided lcd_write (blocking) call. */
-  gpio_num_t reset_pin;        /**< GPIO used for resetting the display. */
+  write_command_fn write_command; /**< Function which the display driver uses to write commands to
+                                     the display. */
+  send_lines_fn lcd_send_lines;   /**< Function which the display driver uses to send bulk (color)
+                     data (non-blocking) to be written to the display. */
+  gpio_num_t reset_pin;           /**< GPIO used for resetting the display. */
```
finger563, line 42: since `reset` and `data_command` can be `GPIO_NUM_NC` (down below), I'd recommend adding default values for both of them to be `GPIO_NUM_NC` in this struct.

finger563: And mark in the comment that the parameter is optional depending on the display driver in question having those pins broken out / functional.

components/display_drivers/include/display_drivers.hpp
```diff
@@ -75,10 +76,11 @@ enum class Flags {
 /**
  * @brief Command structure for initializing the lcd
  */
-struct LcdInitCmd {
-  uint8_t command;  /**< Command byte */
-  uint8_t data[16]; /**< Data bytes */
-  uint8_t length; /**< Number of data bytes; bit 7 means delay after, 0xFF means end of commands. */
+template <typename Command = uint8_t> struct LcdInitCmd {
+  Command command;       /**< Command byte */
+  uint8_t data[16] = {}; /**< Data bytes */
```
finger563, line 81: given that we frequently have command parameters less than 16 and it doesn't seem to make much sense to force the user to add length (which can be error prone), let's go ahead and make this a `std::vector<uint8_t>`. this will still allow the initializer list initialization we're using throughout the code, but we can remove the `uint8_t length` from the command struct and we no longer have to specify the length when we're building the command list.

components/display_drivers/include/gc9a01.hpp
```diff
@@ -332,62 +327,28 @@ class Gc9a01 {
...
-    send_command(Command::ramwr);
     uint32_t size = width * height;
     static constexpr int max_bytes_to_send = 1024 * 2;
     uint16_t color_data[max_bytes_to_send];
     memset(color_data, color, max_bytes_to_send * sizeof(uint16_t));
     for (int i = 0; i < size; i += max_bytes_to_send) {
-      int num_bytes = std::min((int)(size - i), (int)(max_bytes_to_send));
-      send_data((uint8_t *)color_data, num_bytes * 2);
+      int num_bytes = std::min(static_cast<int>(size - i), (int)(max_bytes_to_send));
+      write_command_(static_cast<uint8_t>(Command::ramwr), reinterpret_cast<uint8_t *>(color_data),
```
finger563, line 336: It's a little bit less efficient than the previous code (since previously we'd only send the command once), but I suppose that's OK for this since the bulk of the time will still be spend sending the color data.

Perhaps it would make sense to make the `command` parameter a `std::optional`? so that the `write_command_` function that's provided can handle the case where we don't want to send a new command but instead are just continuing to send more data?

components/display_drivers/include/gc9a01.hpp
```diff
@@ -332,62 +327,28 @@ class Gc9a01 {
...
-   * @param length Number of bytes of data to send.
-   * @param flags Optional (default = 0) flags associated with transfer.
-   */
-  static void send_data(const uint8_t *data, size_t length, uint32_t flags = 0) {
-    flags |= (1 << (int)display_drivers::Flags::DC_LEVEL_BIT);
-    lcd_write_(data, length, flags);
-  }
-
-  static void send_commands(display_drivers::LcdInitCmd *commands) {
+  static void send_commands(std::span<display_drivers::LcdInitCmd<>> commands) {
```
finger563, line 345: might update this to have the arg be `const std::span<display_drivers::LcdInitCmd<> const>& commands` (or more concisely `const std::span<LcdInitCmd const> &commands` once/if you add the subclass + `using` directive.

components/display_drivers/include/gc9a01.hpp
```diff
@@ -332,62 +327,28 @@ class Gc9a01 {
...
-    uint16_t cmd = 0;
-    while (commands[cmd].length != 0xff) {
-      send_command(commands[cmd].command);
-      send_data(commands[cmd].data, commands[cmd].length & 0x1F);
-      if (commands[cmd].length & 0x80) {
-        std::this_thread::sleep_for(100ms);
-      }
-      cmd++;
+
+    for (auto [command, data, length, delay_ms] : commands) {
```
finger563, line 348: prefer `for (const auto &[command, data, length, delay_ms] : commands) {`

components/display_drivers/include/display_drivers.hpp
finger563: The upside of this would be that much of the common config info such as offset, mirror, mutex, etc. can be moved into the parent `DisplayDriver` which greatly improves the code.

components/display_drivers/include/gc9a01.hpp
```diff
@@ -454,17 +415,17 @@ class Gc9a01 {
...
+  static inline gpio_num_t reset_pin_;
+  static inline gpio_num_t dc_pin_;
+  static inline int offset_x_;
+  static inline int offset_y_;
+  static inline bool mirror_x_;
+  static inline bool mirror_y_;
+  static inline bool mirror_portrait_;
+  static inline bool swap_xy_;
+  static inline bool swap_color_order_;
+  static inline std::mutex spi_mutex_;
```
finger563, line 429: See my comment above about moving these into the parent class `espp::DisplayDriver`, which of course means they are no longer static.

components/display_drivers/include/sh8601.hpp
```diff
@@ -0,0 +1,364 @@
+#pragma once
+
+#include <mutex>
+
+#include "display_drivers.hpp"
+
+namespace espp {
+/**
+ * @brief Display driver for the GC9A01 display controller.
```
finger563, line 9: Please update the doxy comment for this class when you're done to have correct info about the driver :)

components/display_drivers/include/display_drivers.hpp
finger563: Note: this approach (while overall a better approach IMO) does have the downside of needing to refactor the `Display` class some more, since right now the rotation callback and the `flush` callback must be static members. 

That design is not a great design, so I'd suggest:
1. Update the display class to take in `std::function` typdefs for rotation callback and flush callbacks.
2. Update the display class to store the rotation callback and flush callback internally, to be called in the static callback handlers.
3. Implement the static flush callback within the display class itself, and then use the `display` pointer passed to the `flush` callback to determine which registered `flush` callback to call.

This has the added benefit of automatically updating the code to be able to support multiple displays more effectively and cleanly (I think).

This sounds like a pretty hefty refactor, so please let me know if there is any part of it you'd like me to tackle within espp first.

finger563: Thinking about it some more, I may just have this be a follow-up PR after this is merged

finger563 (approved):
Changes look good to me! I'll test on a couple of my boards to make sure.

Your changes have pointed out a refactor that I think should be done to this part of the code that I mentioned in some of the comments on this PR. Let me know what you think of the proposed changes and if you have any feedback related to them.

finger563 (changes requested):
Testing some of the example projects on my local devices and the display no longer looks correct (it's mostly garbage). Also I think you need to update the seeed studio example.

components/t-deck/src/t-deck.cpp
```diff
@@ -363,29 +363,43 @@ void IRAM_ATTR TDeck::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 395: this should be `&trans[1]`

components/t-dongle-s3/src/t-dongle-s3.cpp
```diff
@@ -250,29 +250,43 @@ void IRAM_ATTR TDongleS3::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 282: this should be `&trans[1]`

components/wrover-kit/src/wrover-kit.cpp
```diff
@@ -141,29 +141,43 @@ void IRAM_ATTR WroverKit::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 173: this should be `&trans[1]`

components/seeed-studio-round-display/src/seeed-studio-round-display.cpp
```diff
@@ -296,29 +296,43 @@ void IRAM_ATTR SsRoundDisplay::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 328: this should be `&trans[1]`

components/matouch-rotary-display/src/matouch-rotary-display.cpp
```diff
@@ -336,30 +336,43 @@ void IRAM_ATTR MatouchRotaryDisplay::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 368: this should be `&trans[1]`

components/esp-box/src/video.cpp
```diff
@@ -134,29 +134,43 @@ void EspBox::lcd_wait_lines() {
...
+      user_data | (1 << static_cast<int>(display_drivers::Flags::DC_LEVEL_BIT)));
+
+  esp_err_t ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
   if (ret != ESP_OK) {
-    logger_.error("Couldn't queue spi trans for display: {} '{}'", ret, esp_err_to_name(ret));
+    logger_.error("Couldn't queue spi command trans for display: {} '{}'", ret,
+                  esp_err_to_name(ret));
   } else {
-    num_queued_trans++;
+    ret = spi_device_queue_trans(lcd_handle_, &trans[0], 10 / portTICK_PERIOD_MS);
```
finger563, line 166: this should be `&trans[1]`

finger563 (commented):
Locally, I've updated the examples to fix the `&trans[1]` issue, and that seems to have resolved the garbage, so the code seems to be working fine on my end :)

components/display_drivers/include/display_drivers.hpp
finger563: I'm going to resolve this conversation, as I"ll just make this part of a fast-follow PR

components/display_drivers/include/gc9a01.hpp
finger563: Similar to the above, since the code works and is cleaner, I'll resolve this PR and if we want later we can make this a future PR.

components/display_drivers/include/gc9a01.hpp
finger563: again per the comment above, that will be a future PR.

finger563:
Once you fix the examples and address the unresolved comments, please Update the docs (add the new file to the `doc/Doxyfile` alongside the existing display-drivers includes and please add it to the `doc/en/display/display_drivers.rst` alongside the others.

Once those are done, I think this will be good to merge :)

components/display_drivers/include/display_drivers.hpp
finger563: Let me know if you wanna do this as part of this PR or if you wanna just push it to a future PR (since we'll be refactoring things here in a little bit anyway)

components/display/include/display.hpp
abrondijk: Ah this was not intentional

components/display/include/display.hpp
abrondijk: Agreed, also took the opportunity to move the memory allocation to `init_memory` .

components/display_drivers/include/display_drivers.hpp
abrondijk: I agree it is error prone, can't say I haven't fallen into that trap myself a couple times...
It's not that big of a change so I'll include it.

abrondijk: Also renamed it to `DisplayInitCmd` as Lcd is not inclusive anymore ;)

abrondijk:
@finger563 I believe I have implemented all requested changes.
The LcdInitCommands refactor was quite a bit more than I initially thought, so I'm afraid you'll have to retest the changes on your hardware.

finger563:
> @finger563 I believe I have implemented all requested changes.
> The LcdInitCommands refactor was quite a bit more than I initially thought, so I'm afraid you'll have to retest the changes on your hardware.

Thanks @abrondijk ! I'll test on some of my displays soon. 

Can you please update the doc folder accordingly, and if possible, can you get a video or still image of the example running on your encoder board and add it to the PR and the example readme?

abrondijk:
> > @finger563 I believe I have implemented all requested changes.
> > The LcdInitCommands refactor was quite a bit more than I initially thought, so I'm afraid you'll have to retest the changes on your hardware.
> 
> Thanks @abrondijk ! I'll test on some of my displays soon.
> 
> Can you please update the doc folder accordingly, and if possible, can you get a video or still image of the example running on your encoder board and add it to the PR and the example readme?

I'll take a photo and video later today!
Can you give me a hint on what I should update in the doc directory?
I tried building the docs, but after building the docker image I'm getting quite a lot of python errors

finger563:
> > > @finger563 I believe I have implemented all requested changes.
> > > The LcdInitCommands refactor was quite a bit more than I initially thought, so I'm afraid you'll have to retest the changes on your hardware.
> > 
> > Thanks @abrondijk ! I'll test on some of my displays soon.
> > 
> > Can you please update the doc folder accordingly, and if possible, can you get a video or still image of the example running on your encoder board and add it to the PR and the example readme?
> 
> I'll take a photo and video later today!
> Can you give me a hint on what I should update in the doc directory?
> I tried building the docs, but after building the docker image I'm getting quite a lot of python errors

Hey @abrondijk add the new file to the doc/Doxyfile alongside the existing display-drivers includes and please add it to the doc/en/display/display_drivers.rst alongside the others. That should be all that's required. The docs build but produce a lot of errors, at some point I need to go back and do a pass on the docs to clean them up 😅

abrondijk:
@finger563 Is that not what I did [here](https://github.com/esp-cpp/espp/pull/383/commits/74e44c316303320291a3f00b1512b33738fbc649)?

abrondijk:
Sidenote, I will slightly adjust the demo gui, as otherwise the label with the iteration count is barely visible on round displays (like this one).

finger563:
> @finger563 Is that not what I did [here](https://github.com/esp-cpp/espp/pull/383/commits/74e44c316303320291a3f00b1512b33738fbc649)?

Ah yes you did, my bad I missed that 😅

abrondijk:
Pipelines should succeed now, sorry about that.

abrondijk:
@finger563 I took a photo and video, where do you want them in the repo? In the display_driver example directory?

finger563:
> @finger563 I took a photo and video, where do you want them in the repo? In the display_driver example directory?

@abrondijk I usually just upload them into the PR as a GitHub asset so they're not in the repo, then once they're uploaded I copy the markdown it generated into the example readme :)

abrondijk:
Fair enough! Here you go:

https://github.com/user-attachments/assets/fef57074-bab7-4fdf-a25e-f4590773c926

SH8601

finger563:
Tested:
- [x] t-dongle-s3
- [x] esp-box
- [x] t-deck
- [x] seeed-studio-round-display
- [x] matouch-rotary-display

The only thing I don't like is now there are a _lot_ of warnings for `missing initializer for member ...::parameters`. We can remove these warnings by default constructing / setting default for the parameter vector to be `{}`, so please do that. Once that's done I'll merge it 🚀

finger563 (approved):
Just add the default `{}` to the `DisplayInitCmd::parameters`, and this will be good to merge! 

Thanks for doing this!

As mentioned in the previous comments, once this is merged I'll start the work to do a refactor for the display drivers so they inherit from a base `DisplayDriver` class.

abrondijk:
Done!
Thank you as well for your time and patience. We greatly appreciate your library!

As for the upcoming refactor, I have some thoughts/ideas/comments and would of course be happy to help. Would you like them here, or do you want to open an issue/pr for that?

finger563:
> Done!
> Thank you as well for your time and patience. We greatly appreciate your library!
> 
> As for the upcoming refactor, I have some thoughts/ideas/comments and would of course be happy to help. Would you like them here, or do you want to open an issue/pr for that? 

I actually made a discord so that it's easier to have discussions for things like this :) 

Link is in espp readme

PR #400, opened 2025-03-07
feat(display): Remove display task to instead have lvgl handle it directly

abrondijk:
## Description

This change removes the update task entirely from the display component, as lvgl can now retrieve its own task ticks.

It also sets the delay callback, so that it can actually delay, rather than doing [this](https://github.com/lvgl/lvgl/blob/7f07a129e8d77f4984fff8e623fd5be18ff42e74/src/tick/lv_tick.c#L89-L97).
This should be much more efficient.

Potential downside is that the pause and resume functions can no longer do anything.
The user could still "pause" lvgl by not calling `lv_timer_handler()`.

## Motivation and Context
General performance improvements. On programs that actually have a lot of tasks running, the delay callback should free up quite some cpu time.

## How has this been tested?
Tested on the Lilygo T-Encoder-Pro and Smartknob-HA.

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update
- [ ] Software change

## Checklist:

### Software

### Hardware

### DIFF
--- a/components/display/include/display.hpp
+++ b/components/display/include/display.hpp
@@ -387,7 +387,7 @@ protected:
 
     lv_tick_set_cb(xTaskGetTickCount);
 
-    lv_delay_set_cb(vTaskDelay);
+    lv_delay_set_cb(vTaskDelay_wrapper);
 
     display_ = lv_display_create(width_, height_);
     // store a pointer to this object in the display user data, so that we can
--- a/components/display/include/display.hpp
+++ b/components/display/include/display.hpp
@@ -80,12 +80,6 @@ public:
         nullptr}; /**< Optional function used to configure display with new rotation setting. */
     DisplayRotation rotation{
         DisplayRotation::LANDSCAPE}; /**< Default / Initial rotation of the display. */
-    Task::BaseConfig task_config{.name = "Display",
-                                 .stack_size_bytes = 4096,
-                                 .priority = 20,
-                                 .core_id = 0}; /**< Task configuration. */
-    std::chrono::duration<float> update_period{
-        0.01}; /**< How frequently to run the update function. */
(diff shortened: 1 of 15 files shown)


finger563:
I've tested this locally within the `esp-box-emu` project, and simply calling the existing `gui.pause()` and `menu.pause()` (which we had to anyway) is sufficient to allow direct rendering of emulator to screen for direct mode. Can still easily switch back and forth between lvgl and emulator so this works well in my testing 🚀 

Please fix the broken examples, then I think this will be good.

components/display/include/display.hpp
```diff
@@ -479,35 +447,12 @@ template <typename Pixel> class Display : public BaseComponent {
...
-    {
-      using namespace std::chrono_literals;
-      std::unique_lock<std::mutex> lk(m);
-      cv.wait_for(lk, update_period_, [&task_notified] { return task_notified; });
-      task_notified = false;
-    }
-    // don't want to stop the task
-    return false;
-  }
+  static void vTaskDelay_wrapper(const uint32_t ms) { vTaskDelay(pdMS_TO_TICKS(ms)); }
```
finger563, line 454: I'm not seeing this used anywhere, can you remove this?

abrondijk: Huh, I'm surprised it's worked properly, as the lvgl delay callback expects the function to delay in miliseconds, whereas vTaskDelay takes ticks...
That's why this function is there, the callback should have been set to this rather than vTaskDelay directly.

finger563: Probably worked in my tests because I set freertos tick to 1000 hz so it's basically the same thing

PR #447, opened 2025-06-03
chore: Remove unused/outdated version from idf-component files

finger563:


finger563:
Need to keep version string it seems

finger563:
Based on some testing, it seems the errors that I was getting here are not related to removing the version from the manifest file.

In fact the [readme](https://github.com/espressif/upload-components-ci-action?tab=readme-ov-file#usage) even says so:

> If the version of the component is not specified in the manifest file, you can use the version parameter. It must be a valid [component version](https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/versioning.html) optionally prefixed with the character "v". I.e. versions formatted like v1.2.3 or 1.2.3 are supported.

finger563:
https://github.com/espressif/upload-components-ci-action/issues/24

PR #466, opened 2025-07-05
feat(gfps): Added processing to control non-discoverable advertising …

hirokada:
GFPS: Expose AccountKey write and NDA generation events for subsequent pairing integration

### Summary

This PR introduces internal event exposure mechanisms across the GFPS service layer and the embedded Fast Pair library, enabling application-level control during the subsequent pairing phase.

To support external advertisement orchestration, the GFPS layer now receives structured notifications when:

- A remote device writes an Account Key
- A Non-Discoverable Advertisement (NDA) payload is generated

These enhancements enable tight integration with existing application-layer advertisement logic, especially in scenarios where discoverable and non-discoverable flows must be coordinated.

This PR builds upon esp-cpp/nearby#2, now integrated via submodule update.

#### GFPS Layer (Service-level integration for application control)

**Files:**

- `gfps.hpp`
- `gfps_service.hpp`
- `nearby_ble.cpp`

**Enhancements:**

- Added two opt-in `Config` callbacks:
  - `account_key_write_callback_t`
  - `nda_ready_callback_t`
- Removed legacy `extern` hooks for event propagation
- Safely nullified stored callbacks in `gfps::deinit()` to prevent residual behavior
- Refactored passkey injection logic to **target the actual peer initiating pairing**, instead of defaulting to the most recently connected device

#### Embedded Layer (GFPS client callbacks)

**Files:**

- `nearby_fp_client.h`
- `nearby_fp_client.c`

**Enhancements:**

- Added two client-level callback fields to `NearbyFpClientCallbacks`:
  - `on_account_key_written`
  - `on_nondiscoverable_advertisement_ready`
- These callbacks run independently of the legacy `on_event()` dispatch
- Introduced `nearby_fp_client_Deinit()` stub for lifecycle management
- NDA payload is logged for debugging and observability

### Compatibility

| Area                  | Status                                                       |
| --------------------- | ------------------------------------------------------------ |
| Existing apps         | Unaffected (new callbacks default to nullptr; opt-in only)   |
| Legacy embedded logic | Transitioned from on_event() dispatch to direct callback fields |
| Discoverable vs NDA   | Selectively controllable via Config callbacks                |
| Pairing versions      | Compatible with GFPS v2/v3                                   |
| Passkey routing       | Now dynamically targets correct initiator; fallback behavior preserved |

## Motivation and Context

Prior implementations assumed advertising behavior was controlled solely within the GFPS library. However, in real-world applications—especially during subsequent pairing—advertisements must align with runtime platform logic.

By exposing these internal events:

- Apps can delay, suppress, or override NDA advertisement scheduling
- Pairing behavior becomes externally observable and modifiable
- Control parity between initial and subsequent pairing is achieved

Prior GFPS implementations routed passkey confirmation to the last-connected peer, regardless of pairing initiator. This change ensures the passkey is returned to the correct device that initiated the pairing exchange.

## How has this been tested?

- Verified correct emission of `NdaPayloadReady` post-generation

- Confirmed `AccountKeyWritten` is triggered only after key registration completes

- Tested suppression and injection of NDA ads using `Config` callbacks

- Validated fallback behavior (no callbacks set) matches legacy logic

- The following actual devices were used for these checks.

  - Pixel 7 Pro (Android 16)

  - Galaxy S24 (Android 15)

Also verified legacy behavior when callbacks are unset, ensuring non-regression.
## Types of changes

- [ ] Documentation Update
- [x] Software change

## Checklist:

### Software

### DIFF
--- a/components/gfps_service/include/gfps.hpp
+++ b/components/gfps_service/include/gfps.hpp
@@ -46,6 +46,17 @@ typedef std::function<bool(nearby_fp_Characteristic, const uint8_t *, size_t)> n
 /// Configuration for the Google Fast Pair Service
 struct Config {
   notify_callback_t notify; ///< Callback to enable gfps to notify the remote device of changes
+
+  /// Optional callback to handle passkey confirmation during pairing.
+  /// If not provided, GFPS will default to using the first connected peer.
+  /// The callback should call NimBLEDevice::injectConfirmPasskey with the correct peer if available.
+  std::function<void(uint32_t passkey)> set_passkey_callback = nullptr;
+
+  /// Optional callback for when an Account Key is written by the remote device.
+  std::function<void(uint64_t peer_addr, const uint8_t *key)> on_account_key_write_callback = nullptr;
+
+  /// Optional callback triggered when the GFPS layer requests to initiate Non-Discoverable Advertising (NDA).
+  std::function<void(const uint8_t *adv_data, size_t len)> on_non_discoverable_advertisement_ready_callback = nullptr;
 };
 
 /// Initialize the Google Fast Pair Service
--- a/components/gfps_service/include/gfps.hpp
+++ b/components/gfps_service/include/gfps.hpp
@@ -43,6 +43,20 @@ namespace gfps {
 /// @param length The length of the data
 typedef std::function<bool(nearby_fp_Characteristic, const uint8_t *, size_t)> notify_callback_t;
 
+/// Callback for assigning a passkey to the pairing peer
+/// @param passkey The passkey value to be injected
+typedef std::function<void(uint32_t)> set_passkey_callback_t;
+
+/// Callback invoked when an Account Key is written by a remote device
+/// @param peer_addr BLE address of the remote device
+/// @param key Pointer to the 16-byte Account Key data
+typedef std::function<void(uint64_t, const uint8_t *key)> account_key_write_callback_t;
+
+/// Callback triggered when the GFPS layer finishes constructing a Non-Discoverable Advertisement payload
+/// @param adv_data Pointer to raw advertisement payload bytes
+/// @param len Length of the advertisement data
+typedef std::function<void(const uint8_t *adv_data, size_t len)> nda_ready_callback_t;
+
 /// Configuration for the Google Fast Pair Service
 struct Config {
   notify_callback_t notify; ///< Callback to enable gfps to notify the remote device of changes
@@ -50,13 +64,13 @@ struct Config {
   /// Optional callback to handle passkey confirmation during pairing.
   /// If not provided, GFPS will default to using the first connected peer.
   /// The callback should call NimBLEDevice::injectConfirmPasskey with the correct peer if available.
-  std::function<void(uint32_t passkey)> set_passkey_callback = nullptr;
+  set_passkey_callback_t set_passkey_callback = nullptr;
 
   /// Optional callback for when an Account Key is written by the remote device.
-  std::function<void(uint64_t peer_addr, const uint8_t *key)> on_account_key_write_callback = nullptr;
+  account_key_write_callback_t on_account_key_write_callback = nullptr;
 
   /// Optional callback triggered when the GFPS layer requests to initiate Non-Discoverable Advertising (NDA).
-  std::function<void(const uint8_t *adv_data, size_t len)> on_non_discoverable_advertisement_ready_callback = nullptr;
+  nda_ready_callback_t on_non_discoverable_advertisement_ready_callback = nullptr;
 };
 
 /// Initialize the Google Fast Pair Service
--- a/components/gfps_service/include/gfps.hpp
+++ b/components/gfps_service/include/gfps.hpp
@@ -51,7 +51,7 @@ typedef std::function<void(uint32_t)> set_passkey_callback_t;
 /// Callback invoked when an Account Key is written by a remote device
 /// @param peer_addr BLE address of the remote device
 /// @param key Pointer to the 16-byte Account Key data
-typedef std::function<void(uint64_t, const std::span<uint8_t, 16>&)> account_key_write_callback_t;
+typedef std::function<void(uint64_t, const std::span<const uint8_t, 16>&)> account_key_write_callback_t;
 
 /// Callback triggered when the GFPS layer finishes constructing a Non-Discoverable Advertisement payload
 /// @param adv_data Pointer to raw advertisement payload bytes
--- a/components/gfps_service/include/gfps.hpp
+++ b/components/gfps_service/include/gfps.hpp
@@ -4,6 +4,7 @@
 #include <cstdint>
 #include <functional>
 #include <vector>
+#include <span>
 
 #include <sdkconfig.h>
 
@@ -50,12 +51,12 @@ typedef std::function<void(uint32_t)> set_passkey_callback_t;
 /// Callback invoked when an Account Key is written by a remote device
 /// @param peer_addr BLE address of the remote device
 /// @param key Pointer to the 16-byte Account Key data
-typedef std::function<void(uint64_t, const uint8_t *key)> account_key_write_callback_t;
+typedef std::function<void(uint64_t, const std::span<uint8_t, 16>&)> account_key_write_callback_t;
 
 /// Callback triggered when the GFPS layer finishes constructing a Non-Discoverable Advertisement payload
 /// @param adv_data Pointer to raw advertisement payload bytes
 /// @param len Length of the advertisement data
-typedef std::function<void(const uint8_t *adv_data, size_t len)> nda_ready_callback_t;
+typedef std::function<void(const std::span<const uint8_t>&)> nda_ready_callback_t;
 
 /// Configuration for the Google Fast Pair Service
 struct Config {
--- a/components/gfps_service/include/gfps_service.hpp
+++ b/components/gfps_service/include/gfps_service.hpp
@@ -157,6 +157,32 @@ public:
     NimBLEDevice::setCustomGapHandler(gfps::ble_gap_event_handler);
   }
 
+  /// Initialize the service with custom GFPS configuration.
+  /// \param server The BLE server to attach the service to
+  /// \param app_config Application-provided callbacks (e.g., passkey handling, Account Key, NDA advertisement)
+  /// \note Use this overload to inject application-specific behavior into GFPS logic  
+  void init(NimBLEServer *server, const espp::gfps::Config &app_config) {
+    // make the service
+    make_service(server);
+
+    espp::gfps::Config full_config;
+    full_config.notify =
+        [this](nearby_fp_Characteristic characteristic, const uint8_t *value, size_t length) {
+          // notify the characteristic
+          return notify(characteristic, value, length);
+        };
+    full_config.set_passkey_callback = app_config.set_passkey_callback;
+    full_config.on_account_key_write_callback = app_config.on_account_key_write_callback;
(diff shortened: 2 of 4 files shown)


finger563 (changes requested):
Hey @hirokada-san, Thanks for this PR! I don't think we want to have this be a kconfig-setting and we definitely can't have `espp::gfps_service` be dependent on an opaque `app.hpp` and `bb::App` API. 

Instead what I'd suggest is that the `gfps_service` take in another function that the user can provide which would then be called in the `nearby_platform_SetRemotePasskey` implementation, if the function is not `nullptr`. I'd expect the function to be provided in the `espp::gfps::Config` struct, similar to the `notify_callback_t` that's provided today. For backwards compatibility for the API, I'd expect the function to default to nullptr in the struct. If no callback is provided, then the original implementation of the `SetRemotePasskey` function runs (which ends up calling `injectConfirmPasskey` on the first connected peer). If the callback is provided, then it should be the one that handles all the work, and should simply be provided the passkey from the outer call.

I believe that such a design requires no Kconfig setting, no `app.hpp` or other opaque API dependencies, and maximizes flexibility for use of the API in the future. Additionally, it continues to remain backwards compatible because existing callers (who are not providing the function pointer today) will have the same behavior as they do today.

components/gfps_service/src/nearby_ble.cpp
```diff
@@ -10,12 +10,18 @@ static const nearby_platform_BleInterface *g_ble_interface = nullptr;
 static const nearby_platform_BtInterface *g_bt_interface = nullptr;
 
 static espp::gfps::notify_callback_t g_gfps_notify_cb = nullptr;
+static std::function<void(uint32_t)> g_set_passkey_cb = nullptr;
+static std::function<void(uint64_t peer_addr, const uint8_t *key)> g_account_key_write_cb = nullptr;
+static std::function<void(const uint8_t *adv_data, size_t len)> g_non_discoverable_advertisement_cb = nullptr;
```
finger563, line 15: Please use the typedefs that you add here for clarity and consistency

components/gfps_service/include/gfps.hpp
```diff
@@ -46,6 +46,17 @@ typedef std::function<bool(nearby_fp_Characteristic, const uint8_t *, size_t)> n
...
+  /// Optional callback to handle passkey confirmation during pairing.
+  /// If not provided, GFPS will default to using the first connected peer.
+  /// The callback should call NimBLEDevice::injectConfirmPasskey with the correct peer if available.
+  std::function<void(uint32_t passkey)> set_passkey_callback = nullptr;
+
+  /// Optional callback for when an Account Key is written by the remote device.
+  std::function<void(uint64_t peer_addr, const uint8_t *key)> on_account_key_write_callback = nullptr;
+
+  /// Optional callback triggered when the GFPS layer requests to initiate Non-Discoverable Advertising (NDA).
+  std::function<void(const uint8_t *adv_data, size_t len)> on_non_discoverable_advertisement_ready_callback = nullptr;
```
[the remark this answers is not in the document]
hirokada: Thanks for the suggestion — I agree typedefs would make the `Config` struct more consistent and readable.

I've defined:
- `set_passkey_callback_t`
- `account_key_write_callback_t`
- `nda_ready_callback_t`

and updated `gfps::Config` accordingly.

components/gfps_service/include/gfps_service.hpp
```diff
@@ -157,6 +157,32 @@ class GfpsService : public espp::BaseComponent {
     NimBLEDevice::setCustomGapHandler(gfps::ble_gap_event_handler);
   }
 
+  /// Initialize the service with custom GFPS configuration.
+  /// \param server The BLE server to attach the service to
+  /// \param app_config Application-provided callbacks (e.g., passkey handling, Account Key, NDA advertisement)
+  /// \note Use this overload to inject application-specific behavior into GFPS logic  
+  void init(NimBLEServer *server, const espp::gfps::Config &app_config) {
```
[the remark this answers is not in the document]
hirokada: @finger563 -san, great point! I've updated `init(NimBLEServer*)` to delegate directly to the overloaded `init(server, config)` as suggested.

This reduces duplication and keeps the configuration logic centralized. Thanks for catching this!

components/gfps_service/src/nearby_ble.cpp
hirokada: @finger563 -san, absolutely — thanks for pointing that out! 

I've updated these static std::function declarations to use the corresponding typedefs (set_passkey_callback_t, account_key_write_callback_t, and nda_ready_callback_t) for consistency with the Config struct. 

This should make the callback types easier to scan and align with the rest of the codebase.

components/gfps_service/src/nearby_ble.cpp
```diff
@@ -34,6 +40,27 @@ const nearby_platform_BleInterface *espp::gfps::get_ble_interface() { return g_b
...
+}
+// Notifies the application when NDA advertisement data is ready.
+void gfps_on_non_discoverable_advertisement_ready(const uint8_t *adv_data, size_t len) {
+  if (g_non_discoverable_advertisement_cb) {
+    g_non_discoverable_advertisement_cb(adv_data, len);
+  }
+}
+
+}  // extern "C"
+
```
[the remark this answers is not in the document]
hirokada: @finger563 -san, thanks for the feedback. You're absolutely right — the issue wasn’t with the approach itself, but rather that these `extern "C"` functions introduced a cross-library interface without sufficient visibility.

I've since replaced this with structured event dispatch via `client_callbacks->on_event`, which keeps the boundaries between the embedded and GFPS layers clean and explicit. I've removed the obsolete functions to avoid confusion.

Fast Pair Event Notification Flow (Before & After)
- Previous Approach: Notification via extern "C" Hooks
nearby_fp_client.c
├─ Writes Account Key
├─ Generates NDA Payload
│
↓
extern "C" hooks
├─ gfps_on_account_key_write()
└─ gfps_on_non_discoverable_advertisement_ready()
        └─ Used directly by the GFPS layer, but the connection path is opaque due to cross-library boundaries

- Current Approach: Structured Dispatch with Explicit Event Routing
nearby_fp_client.c
├─ Writes Account Key
├─ Generates NDA Payload
│
↓
Structured Dispatch
├─ client_callbacks->on_event(evt)     ◀─ Explicit event type + payload
│
└─ handle_event(evt) in GFPS layer
    └─ Triggers configured app-level callback
       (e.g., on_account_key_write_callback)


These functions were part of an earlier plan to expose embedded-layer events via direct extern "C" hooks. However, I've transitioned to structured dispatch using `client_callbacks->on_event`, which allows us to define clear event types and payloads in `nearby_event.h`.

This approach maintains strong separation between the embedded library and the GFPS service layer, avoids hidden interface dependencies, and improves readability. The legacy externs have been removed accordingly.

components/gfps_service/src/nearby_ble.cpp
finger563: @hirokada I'm not seeing the new typedefs used here

components/gfps_service/include/gfps.hpp
```diff
@@ -43,9 +43,34 @@ namespace gfps {
...
+
+/// Callback invoked when an Account Key is written by a remote device
+/// @param peer_addr BLE address of the remote device
+/// @param key Pointer to the 16-byte Account Key data
+typedef std::function<void(uint64_t, const uint8_t *key)> account_key_write_callback_t;
+
+/// Callback triggered when the GFPS layer finishes constructing a Non-Discoverable Advertisement payload
+/// @param adv_data Pointer to raw advertisement payload bytes
+/// @param len Length of the advertisement data
+typedef std::function<void(const uint8_t *adv_data, size_t len)> nda_ready_callback_t;
```
finger563, line 58: For these two functions, can we change the use of `const uint8_t*` and `const uint8_t*, size_t` to be `const std::span<uint8_t>&`. Given that the first function's argument is known size of 16 bytes, we can use

```c++
const std::span<uint8_t, 16>&
```

While for the second, we'd still use

```c++
const std::span<uint8_t>&
```

This provides a lightweight accessor which enables the application code to use range-based methods and other accessors, while also including OOB checking and other type/memory safety helpers.

Then in the event handler which calls these callbacks you can simply call:

```c++
g_account_key_write_callback_(peer_addr, std::span<uint8_t, 16>(key, 16));
```

and

```c++
g_nda_ready_callback_(std::span<uint8_t>(adv_data, len));
```

components/gfps_service/src/nearby_ble.cpp
```diff
@@ -10,24 +10,66 @@ static const nearby_platform_BleInterface *g_ble_interface = nullptr;
...
+// -----------------------------------------------------------------------------
+static void handle_event(const nearby_event_Event* evt) {
+  switch (evt->event_type) {
+    case kNearbyEventAccountKeyWritten: {
+      auto* payload = reinterpret_cast<const nearby_event_AccountKeyWritten*>(evt->payload);
+      if (g_account_key_write_cb)
+        g_account_key_write_cb(payload->peer_address, payload->key_data);
+      break;
+    }
+    case kNearbyEventNdaPayloadReady: {
```
finger563, line 28: @hirokada the build is failing because these events are not defined. The only events that are defined are:

```c
// Types of events that can be emitted by Nearby library
typedef enum {
  // A client connected over RFCOMM to the message stream endpoint defined in
  // Fast Pair specification
  // Payload: |nearby_event_MessageStreamConnected|
  kNearbyEventMessageStreamConnected = 0,
  // An RFCOMM client disconnected
  // Payload: |nearby_event_MessageStreamDisonnected|
  kNearbyEventMessageStreamDisconnected,
  // A client has sent a message over the message stream. Some incoming messages
  // are consumed by the Nearby library, in which case the event is not emitted.
  // Payload: |nearby_event_MessageStreamReceived|
  kNearbyEventMessageStreamReceived
} nearby_event_Type;
```

finger563 (changes requested):
@hirokada There are still a few lingering issues - one is relatively minor (the use of the new typedef), but the other breaks the build (the use of non-existent event enums). Does this require a fork of the google/nearby library to work or where are these events coming from?

I've also suggested an API update to use `std::span` where possible to do a lightweight type-conversion when passing data out of the low-level gfps library to app c++ code.

components/gfps_service/src/nearby_ble.cpp
hirokada: @finger563 -san, thank you — you're right, those event types come from a patched version of google/nearby we're using internally. To avoid build problems in environments that do not include them, I will modify the event handlers to be built conditionally using #ifdef.

components/gfps_service/src/nearby_ble.cpp
hirokada: @finger563 -san, thank you for pointing that out. I missed that correction. I will fix it in the next commit.

components/gfps_service/src/nearby_ble.cpp
finger563: @hirokada If those functions are necessary, then I'll fork `google/nearby` into `esp-cpp/nearby` so we can use it here and the changes can be tracked + maintained.

hirokada:
@finger563 -san, I’ve applied the suggestions from Copilot and confirmed that the modified source behaves correctly.

finger563 (approved):

finger563:
> @finger563 -san, I’ve applied the suggestions from Copilot and confirmed that the modified source behaves correctly.

@hirokada -san, thank you for all the work you put into reviewing and testing this!

PR #673, opened 2026-07-22
feat: Add new `thread_pool` component built on top of `Task`

guo-max:
## Description

add Thread_pool component

## Motivation and Context

A thread pool that dispatches submitted jobs to a fixed set of worker threads.

## How has this been tested?

an example is added and tested. 
The example was tested on an esp32 board. it use a thread_pool with 2 works and finished 8 jobs.

image

## Types of changes

- [X] Documentation Update
- [X] Software change

## Checklist:

- [X] I have added / updated the documentation related to this change via either README or WIKI

### Software

- [X] I have added tests to cover my changes.

### Hardware


components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    stopping_ = true;
+  }
+
+  queue_has_work_cv_.notify_all();
+  queue_has_space_cv_.notify_all();
+
+  for (auto &worker : workers_) {
+    worker->stop();
+  }
```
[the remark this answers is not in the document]
guo-max: It is ok to not execute the queued jobs. will fix the comments

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    stopping_ = false;
+  }
+
+  for (auto &worker : workers_) {
+    worker->start();
+  }
```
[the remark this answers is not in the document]
guo-max: fixed by logging errors.

components/thread_pool/example/main/thread_pool_example.cpp
finger563, line 1: @guo-max can you please update the example to exercise / demonstrate all of the public APIs of the thread pool class?

components/thread_pool/include/thread_pool.hpp
```diff
@@ -0,0 +1,120 @@
...
+public:
+  /// @brief A callable job that can be submitted to the pool.
+  using Job = std::function<void(void)>;
+
+  /// @brief Snapshot of pool activity counters.
+  struct Stats {
+    std::uint64_t submitted = 0; ///< Total jobs accepted into the queue.
+    std::uint64_t executed = 0;  ///< Total jobs successfully executed.
+    std::uint64_t rejected = 0;  ///< Total jobs rejected (invalid job, stopped/stopping, or queue full).
+  };
```
finger563, line 56: Can you please add format helpers header for this struct so that it can be simply passed to fmt via `{}` (optionally with some format args to control short versus verbose printing)? There are other examples of `*_format_helpers.hpp` within espp that you can use for reference

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  if (config_.max_queue_size > 0) {
+    queue_has_space_cv_.notify_one();
+  }
+
+  job();
+  executed_++;
+
+  return false;
```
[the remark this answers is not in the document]
guo-max: we do not use exceptions. do not change here

components/thread_pool/example/README.md
finger563, line 1: Might update this readme to indicate that it runs a test suite of different usage patterns and API checks ranging from simple usage to more complex / multi-pool usage. Key point is that the end of the example test result is critical for validating functionality of the class.

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,174 @@
...
+    running_.store(true);
+  }
+  return true;
+}
+
+void ThreadPool::stop() {
+  std::lock_guard<std::mutex> lifecycle_lock(lifecycle_mutex_);
+
+  {
+    std::lock_guard<std::mutex> lock(queue_mutex_);
```
finger563, line 72: technically you should be able to move the mutex lock to after the `running_` check/mutate since `running_` is atomic, and the `queue_mutex_` is only protecting the `queue_` call below. Minor nit, so no worries about keeping it / changing it.

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  for (auto &worker : workers_) {
+    worker->stop();
+  }
+}
+
+bool ThreadPool::is_running() const { return running_.load(); }
+
+bool ThreadPool::submit(Job job) {
+  return submit_impl(std::move(job), config_.block_on_submit_when_full);
```
finger563: IIUC since you're using move semantics here, you can actually pass by r-value using `Job &&job` as the parameter signature

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+bool ThreadPool::try_submit(Job job) { return submit_impl(std::move(job), false); }
+
+bool ThreadPool::submit_impl(Job job, bool allow_blocking_when_full) {
+  if (!job) {
+    rejected_++;
+    return false;
+  }
+
+  std::unique_lock<std::mutex> lock(queue_mutex_);
```
finger563, line 99: similarly here, i don't think you need to take the `queue_mutex_` here as early since at this point you're only checking the `atomic` state variables.

guo-max: I will update this and also change the `stopping` to atomic, so `queue_mutex_` only protect queue access.

components/thread_pool/src/thread_pool.cpp
```diff
@@ -0,0 +1,164 @@
...
+  return queue_.size();
+}
+
+std::size_t ThreadPool::worker_count() const { return workers_.size(); }
+
+ThreadPool::Stats ThreadPool::stats() const {
+  return {
+      .submitted = submitted_.load(),
+      .executed = executed_.load(),
+      .rejected = rejected_.load(),
```
finger563, line 138: technically a possible data race here - is this what the `lifecycle_mutex_` is supposed to protect?

It's not really a big issue since the main result is that these are not synchronized with each other, but worth being aware of

finger563 (approved):
Looking good, thanks for making this component! A minor update I might suggest is that the first / primary code snippet / example code for the class should be standard use (not one of the tests) so that it's easier for others to read / follow how they should expect to use it and it's easier for them to copy into their code as a starting point. For that purpose, I might suggest adding another section in the example which is simple use showing with straightforward code, and that section is the first snippet / example in the docstring.

components/thread_pool/src/thread_pool.cpp
guo-max: Not really, `lifecycle_mutex_` is to protect `start/stop` called by differente thread. Otherwise, `start() == True` but it could actually stopped by other thread in the middle of the start.
I agree there maybe a data race here. The result could be the `total job != submitted + rejected` or `submitted != executed + queued`  at a moment. And the numbers should match later.

PR #709, opened 2026-08-12
feat(rtps_embedded): Typed Publisher<T>/Subscriber<T> pub/sub (Phase 5)

finger563:
Phase 5 (typed pub/sub) of the `components/rtps_embedded` refactor — see `REFACTOR_PLAN.md`. Independent of and parallel to the Phase 4 (limits) work.

## What

A header-only **typed** layer over `espp::RtpsParticipant` so applications publish/receive reflectable message structs directly — no manual CDR (de)serialization or byte-span handling.

```cpp
struct Imu { float ax, ay, az; };   // any reflectable struct — no base class/macros

espp::Publisher<Imu>  pub(participant, {.topic="rt/imu", .type_name="sensor_msgs::msg::dds_::Imu_",
                                        .reliability = Reliability::RELIABLE});
pub.publish(Imu{0.1f, 0.2f, 9.8f});

espp::Subscriber<Imu> sub(participant, {.topic="rt/imu", .type_name="sensor_msgs::msg::dds_::Imu_",
                                        .on_message=[](const Imu& m){ use(m); }});
```

- `RtpsMessage` concept: any struct the reflection `cdr` can (de)serialize qualifies.
- `Publisher<T>` serializes with `cdr::serialize_into` into a **reused buffer** → zero steady-state allocation.
- `espp::ros2::topic_name()` helper for ROS 2 naming.

## Additive by design

The untyped span/byte API (`add_writer`/`add_reader`/`publish`/`on_sample`) is unchanged and is exactly what the typed layer wraps. Existing code is unaffected; the golden wire tests are byte-identical.

The esp32 example is converted to the typed API (cleaner than the manual-`cdr::serialize` version, and it verifies the header compiles with reflect-cpp on target). Python keeps the byte API — templates don't bind generically, so the typed layer is a C++ ergonomic layer.

## Verification

- `pc/tests/rtps_typed_pubsub`: two participants exchange a multi-field struct (`uint32` + `float` + `string`) end-to-end over real SEDP/DATA, with **field-level** verification of the typed round-trip — PASS.
- Golden wire tests byte-identical.
- esp32 example builds for `esp32p4` on ESP-IDF 6.0; host lib builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ managed_components/
 
 # docker interop harness build tree (bind-mounted, built in-container)
 pc/build-linux/
+python/env/
--- a/components/rtps_embedded/example/main/main.cpp
+++ b/components/rtps_embedded/example/main/main.cpp
@@ -52,7 +52,10 @@ extern "C" void app_main(void) {
   constexpr const char *sub_topic = "pc_to_mcu";
   constexpr const char *type_name = "std_msgs::msg::String";
 
-  static espp::RtpsParticipant participant({
+  // Automatic locals: they RAII-clean up in reverse order on any early return
+  // (subscriber/publisher stop referencing the participant before it is
+  // destroyed), and the trailing while(true) keeps them alive in normal use.
+  espp::RtpsParticipant participant({
       .interface_address = interface_address,
       .on_publisher_matched = [&]() { logger.info("publisher matched a remote reader"); },
       .on_subscriber_matched = [&]() { logger.info("subscriber matched a remote writer"); },
@@ -66,13 +69,13 @@ extern "C" void app_main(void) {
   // Typed reliable publisher: publish StringMsg structs directly (HEARTBEAT/
   // ACKNACK-acknowledged, retransmitted to matched readers). No manual CDR.
   using Reliability = espp::RtpsParticipant::Reliability;
-  static espp::Publisher<StringMsg> publisher(participant, {
-                                                               .topic = pub_topic,
-                                                               .type_name = type_name,
-                                                               .reliability = Reliability::RELIABLE,
-                                                           });
+  espp::Publisher<StringMsg> publisher(participant, {
+                                                        .topic = pub_topic,
+                                                        .type_name = type_name,
+                                                        .reliability = Reliability::RELIABLE,
+                                                    });
   // Typed subscriber: receive StringMsg structs directly.
-  static espp::Subscriber<StringMsg> subscriber(
+  espp::Subscriber<StringMsg> subscriber(
       participant, {
                        .topic = sub_topic,
                        .type_name = type_name,
@@ -84,7 +87,7 @@ extern "C" void app_main(void) {
   }
 
   // Publish a counter periodically via the typed publisher.
-  static uint32_t counter = 0;
+  uint32_t counter = 0;
   espp::Timer publish_timer({
       .name = "rtps_pub",
       .period = std::chrono::milliseconds(CONFIG_RTPS_EXAMPLE_ANNOUNCE_PERIOD_MS),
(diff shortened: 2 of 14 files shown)


finger563:
Addressed all review comments + the static_analysis failure in 833bb2dc:

- **Missing standard headers** (Copilot): added `<concepts>`/`<cstddef>`/`<memory>`/`<mutex>` — the concept (`std::convertible_to`) and `std::byte`/`std::size_t` were only reachable transitively.
- **Subscriber use-after-free** (Copilot): the `on_sample` callback now captures a `shared_ptr` copy of the user callback instead of `this`, so destroying a `Subscriber` while the participant is still running can't dangle the engine-invoked callback. Documented the lifetime contract (stop the participant before tearing down state the callback references).
- **Publisher data race** (Copilot): `publish()` now guards the reused serialization buffer with a mutex, so concurrent calls from multiple threads are safe.
- **static_analysis** (cppcheck): the findings were in the test — `Telemetry` members now default-initialized (`uninitMemberVarNoCtor`), and the verification loop uses `std::any_of` (`useStlAlgorithm`).

Verified: cppcheck clean on both files with the CI args; typed test passes (sent=5 received=5 with field-level verification); golden byte-identical.

finger563:
Addressed the second review round + added the Python typed layer (f75157a3):

**Review comments (round 2):**
- **strict-aliasing `reinterpret_cast`** (Copilot, both directions): replaced with the standard, well-defined span conversions — `std::as_writable_bytes` on the publish path (`buffer_` is now `std::vector<uint8_t>`) and `std::as_bytes` on the receive path. No `reinterpret_cast` remains.
- **example `static` lifetime** (Copilot): `participant`/`publisher`/`subscriber` are now automatic locals, so they RAII-clean up in the correct order on any early return.

**Python typed pub/sub** (answering the "what's the Python equivalent?" discussion): since `Publisher<T>`/`Subscriber<T>` are C++ templates and can't be bound generically, the Python counterpart is a small pure-Python wrapper — `espp.rtps.Publisher` / `Subscriber` — over the byte API paired with a runtime CDR codec. It's codec-agnostic (any type with `.serialize()`/`.deserialize()`, or explicit callables); [pycdr2](https://pypi.org/project/pycdr2/) dataclasses satisfy it and emit ROS 2 / DDS-compatible CDR. The python examples now publish/receive a complex nested message (`sensor_msgs/msg/Imu`-style: string + nested struct + fixed array + sequence), pycdr2 added to `python/requirements.txt`, and `rtps_typed_test.py` verifies the fields round-trip end-to-end.

Verified: cppcheck clean; C++ + Python typed tests pass (fields verified); golden byte-identical; esp32 example builds.

finger563:
Addressed the new comments in 68fbf9bb:

- **Payload > 64 KB silently truncated** (Copilot / raised for the parallel limits work): `RtpsParticipant::publish` cast the length to `DataSize_t` (`uint16_t`), so oversized samples wrapped. Added `RtpsParticipant::max_payload_size` (65535, `static_assert`'d against the engine's `DataSize_t`) and now **reject** oversized payloads with a logged error in the untyped `publish()` (protects all callers); `Publisher<T>::publish` also early-rejects before serializing. This 64 KB cap is inherent to the RTPS wire format — the DATA submessage `octetsToNextHeader` is 16-bit — so raising it (for images/large data on hosts) is a real protocol feature (DATA_FRAG fragmentation + widening the internal size type), tracked for the limits/refactor track rather than this PR.
- **Demo message reused the official `sensor_msgs/msg/Imu` type name** with a mismatched layout (Copilot): real DDS/ROS 2 peers would match on the name and misdecode. Renamed to demo type names (`espp_examples/msg/Vector3`, `espp_examples/msg/SensorSample`) with a note not to reuse official ROS 2 type names unless the schema matches exactly.

The earlier aliasing / `static` / data-race / include / UAF comments were already resolved in 833bb2dc and f75157a3 (Copilot re-posts them on the pre-fix diff positions). Verified: C++ + Python typed tests pass (fields verified), golden byte-identical, cppcheck clean, esp32 builds.

PR #711, opened 2026-08-14
feat(rtps_embedded): ROS 2-interoperable services (RMI) + actions (AMI) + native protocol

finger563:
Adds **services (RMI)** and **actions (AMI)** to the `rtps_embedded` component, in
both a ROS 2-interoperable flavour and a lean native (espp↔espp) flavour, on top
of the existing pub/sub — validated end-to-end against live ROS 2 (rmw_fastrtps,
Jazzy) both directions.

## What's new

- **Services (RMI)** — request/reply with correlated responses. Byte-level
  (`add_service_server`/`_client`) and typed (`espp::ServiceServer<Req,Resp>` /
  `ServiceClient<Req,Resp>`). Three client call styles: blocking `call`,
  `call_async`, `call_future`. Deferred-reply support (`ServiceResponder`) for
  slow responses.
- **Actions (AMI)** — long-running goals with feedback, result, and cancel
  (3 services + 2 topics; no new wire primitive). Byte-level and typed
  (`ActionServer` / `ActionClient` / `ActionGoalHandle`).
- **Native protocol** — a deliberately lean espp↔espp variant (20-byte in-band
  correlation header, `es_rq`/`es_rr` topics, ~3 endpoints per action vs ROS's
  ~10), same typed wrappers via a `RtpsProtocol` config field.
- **Typed layer** — `rtps_service.hpp` / `rtps_action.hpp`, the RMI/AMI analogue
  of `Publisher<T>`/`Subscriber<T>`: reflectable structs (de)serialized by the
  `cdr` component, no manual bytes.
- **Python bindings** for the whole surface + `python/rtps_rpc_demo.py`.
- **Docs** — `doc/en/protocols/rtps_rmi_ami.rst` (function, use, rationale, wire
  format, when to choose native); refreshed component + example READMEs.
- **Kconfig `RTPS_ENABLE_RPC`** (default on) compiles the whole services/actions
  layer out for pure-pub/sub devices.
- **Embedded example** now hosts a typed service + action a ROS 2 client can
  drive, runs service/action clients, and has an opt-in second-participant
  self-test.

## Wire format (confirmed by capture, not guessed)

Before writing code, each wire format was captured from a live rmw_fastrtps
exchange (tshark) and codified in golden byte tests — which corrected two initial
assumptions: ROS correlation is `related_sample_identity` inline QoS under **both**
PID `0x0083` (OMG) and `0x800f` (eProsima), present on **both** request and reply
(not the `0x8002`/request-only guess). Result: no SEDP type-hash is needed — an
espp service/action even appears in `ros2 service list` / `ros2 action list`.

## Testing

- **Golden** wire tests stay byte-identical; a new `data_related_sample_identity`
  section (all prior bytes unchanged).
- **Interop harness 27/27** (dockerised, gated in CI on `rtps_embedded/**`):
  in-process loopbacks (pub/sub, services, actions, native, typed), byte-level
  codec unit tests, live `ros2 service call` / `ros2 action send_goal` both
  directions, and a functional **Python demo** (5/5) built in-container.
- **esp32** builds green: RPC on and off, static/dynamic storage, frag on/off,
  and the example (now in the CI build matrix).
- Wheel smoke test asserts the full RMI/AMI binding surface.

## Notes for reviewers

- The engine wire change is limited to emitting/parsing the reply inline-QoS
  correlation; **plain pub/sub is byte-identical** (proven by golden + interop).
- A latent portability bug surfaced only when building the Python module on macOS
  libc++ (forward-declared context in a `unique_ptr` member) — fixed; a good
  argument for keeping the wheel build in the review loop.
- The example was renamed `main.cpp` → `rtps_embedded_example.cpp` to prepare for
  retiring the old `rtps` component in a follow-up.
- **Deferred (intentional):** the "shared L2" template-unification of the ROS and
  native paths — the two already share the pub/sub + service primitives; further
  merging is cosmetic and higher-risk.

Design + milestone detail: `components/rtps_embedded/RMI_AMI_DESIGN.md`.


finger563:
Addressed the review comments and static-analysis findings in 8d500399.

**Static analysis (cppcheck):**
- `rtps_action_types.cpp` — replaced the `hex()`+`push_back` loops with an `encap_hex` helper using `insert()` (useStlAlgorithm ×7).
- `rtps_service_loopback.cpp` — renamed the inner deferred-reply `reply` → `dreply` (shadowVariable).
- `rtps_participant.cpp` — `generate_goal_id` now uses `std::generate` (useStlAlgorithm).
- `rtps_service.hpp` — `ServiceClient::call_async` takes `on_response` by const ref (passedByValue).

**Review comments:**
- **`assert` in the demo** — replaced with an explicit check + stderr + non-zero return (asserts are skipped under `python -O`).
- **Missing `canceled()` binding** — exposed `ActionGoalHandle.canceled()` in Python (+ the typed `espp.rtps.GoalHandle`). `NativeGoalHandle` correctly has none (the native protocol has no cancel).
- **Missing `call_future` in Python** — added `call_future()` to the `ServiceClient` and `NativeServiceClient` bindings (returns a `concurrent.futures.Future`) and to the typed `espp.rtps.ServiceClient`; the demo now exercises it (`ros_service_future`) and the wheel smoke test asserts the new methods exist. So the Python surface now matches the docs' claim.
- **Empty `except`** in `espp/rtps.py` — commented (a malformed feedback / raising callback must not kill the engine reader thread).
- **Action result 8-byte alignment** — documented the byte-splice envelope limitation in the user-facing RMI/AMI docs (`get_result`'s result first field must be ≤4-byte aligned; reorder an `int64`/`float64` first field). Goal/feedback payloads and services are unaffected (they follow the 8-aligned UUID / have no such splice). Noted as a v1 limitation to be revisited with full-CDR envelope building.

**Gate:** interop **27/27** (golden byte-identical, refactored `action_types`, `service_loopback`, python demo **6/6** incl. `call_future`); esp32 build green; wheel smoke test green.

finger563:
Addressed the second review batch in c3742371 — these were all async-RPC lifetime/leak issues, so I reworked the teardown and resource ownership coherently. Full gate is green (golden byte-for-byte unchanged, interop **27/27**, native/self loopbacks + `python_rpc_demo` incl. `stop()` all pass).

**`rtps_participant.cpp:351` — deferred responder use-after-free.** Added a shared `Liveness{mutex, alive}` token. `ServiceResponder::reply()` now writes only while holding that mutex with `alive == true`; `stop()` flips `alive = false` **under the same mutex** before the domain/writers are destroyed. A reply that races shutdown either completes against a still-valid writer or observes `!alive` and drops — never a freed writer. The token is a `shared_ptr`, so it outlives the participant.

**`rtps_participant.cpp:1337` — detached goal threads outliving `stop()`.** Execute workers are now **owned** (not detached), stored per server context, reaped as they finish, and **joined in `stop()`** before the domain is torn down. `stop()` became a 4-phase teardown so the join happens *without* `mutex_` held (a worker's final `publish()`/reply must be able to take `mutex_`/`live_` and run to a no-op so it can exit). Accepted goals get a cancel signal first, so cooperative callbacks wind down — execute callbacks are expected to be finite / `is_canceling()`-aware, which I documented.

**`rtps_participant.cpp:696` — goals↔server cycle + unbounded map.** The goal `State` now holds a **`weak_ptr`** back to the server and copies its feedback/status topic names (so publishing needs no server lock). Goals are erased from the server map once their result is delivered (either `get_result` answering a finished goal, or `terminate()` fulfilling a waiting `get_result`), bounding growth. A still-running worker keeps the `State` alive via its own reference, so the erase is safe.

**`rtps_participant.cpp:1213` — native action pre-registration race.** The native client now **buffers** feedback/result that arrives for an unknown handle (bounded at 64 messages) and **replays** it when `send_goal`'s reply installs the goal. A fast native action that publishes before its goal reply lands no longer drops early messages.

**`MessageFactory.hpp:180` — RPC + fragmentation loses correlation.** A `related_sample_identity` DATA carries 60 bytes of inline QoS and **cannot** be fragmented (DATA_FRAG has no inline QoS → correlation lost → timeout). Added `MAX_UNFRAGMENTED_RPC_PAYLOAD = MAX_UNFRAGMENTED_PAYLOAD − 60`; the facade now rejects oversized service request/reply, and both `StatefulWriter`/`StatelessWriter` refuse to fragment an RSI change (drop rather than emit an uncorrelated fragment). Plain (non-RSI) samples still fragment normally — `cross_process_frag_200k` still passes.

**`rtps_embedded_example.cpp:206` — client demo leak with no peer.** The demo now caps the action to **one in-flight goal** (an atomic flag reset on result), so an absent peer can no longer leak a pending goal per 5 s tick. The service `call()` already self-cleans on its 1 s timeout.

PR #714, opened 2026-08-15
feat(socket): generalize Windows guards + add socket-option API

finger563:
## What & why

Two improvements to the `socket` component:

**(a) `_MSC_VER` → `_WIN32` guards.** The Windows-specific code (Winsock2, `closesocket` vs `close`, `WSAStartup`, `INVALID_SOCKET`, `ioctlsocket`, `SO_*`/`char*` casts, `WSAGetLastError`) was guarded on `_MSC_VER`, which only covers the MSVC compiler. Switched every such guard to `_WIN32` so MinGW / clang-cl Windows builds are covered too. Each guard was verified to be Windows-*platform* intent — none were MSVC-compiler-only pragmas.

Files: `socket.hpp`/`socket.cpp`, `tcp_socket.hpp`, `socket_reactor.cpp`, `udp_socket.cpp`, `socket_msvc.hpp`.

**(b) Socket-option setters on the base `Socket`.** Previously, setting `SO_RCVBUF`/`SO_SNDBUF` meant calling `setsockopt()` on the raw native handle (as our RTPS transport did). Added to `espp::Socket`:

```cpp
bool set_option(int level, int option_name, const void *value, size_t size);
template <typename T> bool set_option(int level, int option_name, const T &value);
bool set_receive_buffer_size(size_t bytes);      // SO_RCVBUF
bool set_send_buffer_size(size_t bytes);          // SO_SNDBUF
bool set_reuse_address(bool enable);              // SO_REUSEADDR
std::optional<size_t> get_receive_buffer_size();  // getsockopt (bonus)
```

`set_option` handles the Windows `const char*` `setsockopt` cast and logs on failure. Then migrated the RTPS transport's manual 4 MB `SO_RCVBUF` `setsockopt` (`EsppTransport.cpp`) to `socket->set_receive_buffer_size(...)` — the only such call, and the `Socket` object was in scope.

## Verification

Host lib builds clean (`cmake -S lib ... --target install`, exit 0) — this compiles both the socket sources and the RTPS transport (fragmentation on in the host profile, so the migrated path is exercised). The socket example is ESP-IDF-only, so the host lib is the compile check.


finger563:
Addressed both new Copilot comments in 623f362b:
- **`lib/include/espp.hpp`**: the `windows.h`/`wcswidth` include is now gated on `_WIN32` (not `_MSC_VER`), so MinGW/clang-cl Windows builds pull it in too — consistent with the rest of this PR.
- **`socket.cpp` `get_receive_buffer_size`**: Winsock`getsockopt` takes the optlen as `int*`, not `socklen_t*` — now `int` on `_WIN32`, `socklen_t` elsewhere.

Host lib builds clean.

PR #752, opened 2026-09-02
feat(canopen): DS402 drive panel web app (in-browser SDO over the CAN bridge)

finger563:
A purpose-built, single-file, offline web app for commissioning **CiA 402 (DS402)** drives through the existing USB<->CAN bridge (`can_bridge_example`), using **architecture option A**: all CANopen/DS402 logic runs **in the browser** over raw CAN — the firmware stays a dumb raw-CAN pipe, so **no firmware change**.

Reuses the CAN console's transport + `stream_frame` framing; adds:

- **In-browser CANopen SDO client** — expedited + segmented upload (read) and download (write) on the default SDO channel (`0x600`/`0x580` + nodeId), abort-code decoding, one-transaction-at-a-time.
- **DS402 panel** — statusword (`0x6041`) → power-drive-system state decode; controlword (`0x6040`) command buttons + enable sequence / quick-stop / fault-reset; mode of operation (`0x6060`/`0x6061`); target & actual velocity/position/torque; opt-in live poll.
- **Node identity** (`0x1000`/`0x1008`/`0x1018`) and **NMT** controls.
- **Generic object-dictionary panel** — read/write **any** index:subindex with a chosen type (`u8`..`i32`/string/raw hex), for **vendor-specific** objects.

Since it drives the node over SDO (not cyclic PDOs) it is a commissioning / bring-up tool; the bus must be in **Normal** mode so the bridge ACKs the node.

### Notes
- Auto-listed in the apps index from its `<title>`/`<meta description>`; `canopen.rst` links it beside the CAN console.
- **Verified:** `node --check` clean, plus a runtime test of the SDO expedited/segmented up/download codec + DS402 state decode against a mock CANopen server (13/13 round-trips). Hardware test against a real DS402 drive still pending.
- **Stacked on #748** (needs the `can_bridge_example` firmware). Base will retarget to `main` once #748 merges.
- Possible follow-ups: EDS-file loading for a full object browser; segmented-download is implemented but only lightly exercised.

### DIFF
--- a/components/canopen/web/ds402_panel.html
+++ b/components/canopen/web/ds402_panel.html
@@ -469,6 +469,7 @@
     const STATUS_POLL_MS = 1000;
     const DS402_POLL_MS = 400;
     const SDO_TIMEOUT_MS = 800;
+    const SDO_MAX_TRANSFER = 64 * 1024;   // bound segmented uploads so a faulty node cannot grow memory unbounded
 
     // CANopen COB-IDs (11-bit, function code | node id)
     const COB_NMT = 0x000;
@@ -740,7 +741,10 @@
         try {
           while (this.reading) {
             const { value, done } = await this.reader.read();
-            if (done) break;
+            // A clean EOF while we still expect data (port closed under us, device
+            // unplugged) is a link failure - throw so the caller tears the UI down.
+            // During close() this.reading is already false, so that is not an error.
+            if (done) { if (this.reading) throw new Error("serial connection closed (EOF)"); break; }
             if (value && value.length) onData(new Uint8Array(value.buffer, value.byteOffset, value.byteLength));
           }
         } finally { try { this.reader.releaseLock(); } catch (_) {} this.reader = null; }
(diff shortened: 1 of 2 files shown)


finger563:
Addressed the review comments in `d26ba7bde`:

- **SDO channel not cleaned up on disconnect** — added `abortSdo()` (clears the waiter timer, rejects the pending Promise, resets `sdoTxnActive`), called from both `disconnect()` and `onLinkLost()`. A link drop mid-SDO no longer wedges the channel (`SDO channel busy`) after reconnect or leaves a Promise pending to timeout.
- **Segmented download toggle not validated** — now checks the server echoed the expected toggle bit (matching the upload path) and throws on mismatch, guarding against silent write corruption.
- **`refreshDs402()` re-entrancy** — wrapped the whole multi-read sequence in a `refreshInFlight` guard, so a poll tick or manual refresh cannot start a second refresh mid-sequence and interleave reads (`sdoTxnActive` is released between the individual reads). Manual refresh button simplified accordingly.
- **Misleading NMT node-0 comment** — clarified that the UI always targets the selected node (1–127, shared with the SDO client) and does not expose CANopen node-0 broadcast.

Verified: `node --check` clean, and the SDO codec round-trip test still passes (segmented upload + download with the new toggle check).

finger563:
Addressed the review comments (and rebased onto current `main`):

- **Controls did not reflect bus run-state** — `setStatusLine()` now disables **Apply** + **Start** while the bus is running (the firmware rejects `SET_CONFIG` then) and disables **Stop** while stopped, matching the CAN bridge console. No more clickable "Apply config" during a run despite its "only while stopped" tooltip.
- **Per-frame DOM logging could freeze the UI** — `onIncoming()` logged every non-SDO CAN frame unconditionally; on a live bus (heartbeats / PDOs / EMCY) that floods the DOM. Gated behind a new **"log bus frames"** toggle (default off), mirroring the SDO-frame toggle.
- Also updated the `can_console.html` references to `can_bridge_console.html` (renamed on main in #753) and rebased so the doc link + rename are consistent.

`node --check` clean.

finger563:
Addressed the SDO/DS402 correctness review (`7a8f57715`):

- **SDO response matching** — `sdoDeliver()` now requires an **exact 8-byte, non-RTR, non-extended** frame, and for phases whose response echoes the multiplexer (expedited/initiate/abort) it requires **index:subindex to match** the pending transfer. A late response for an earlier index can no longer complete the next transaction with stale data / a false ack, and stray traffic on the response COB-ID is no longer zero-padded into a fake response. Segment phases (no multiplexer) stay matched by phase + toggle.
- **Node snapshot** — `withSdo()` captures the node id **once** and threads it through the whole transaction, so a node-field change mid segmented transfer cannot split the initiate and segments across servers.
- **Truncated upload** — a segmented upload now **rejects** a short transfer (declared size > bytes delivered) instead of returning truncated data as success.
- **Scalar range / width** — writes are **range-validated** before encoding (`u8=-1`, `i16=65535`, oversized 32-bit no longer silently wrap); reads **reject an undersized object** instead of zero-extending it; **zero-length writes rejected**; target/controlword handlers only log the value as sent when the write actually succeeded.
- **Enable sequence** — the "Enable operation" button now performs the real **state-aware sequence** (Shutdown → Switch On → Enable Operation, with fault reset and quick-stop exit), polling the statusword between transitions, instead of blindly writing `0x000F` (which fails from Switch-on-disabled). Relabeled **"Enable sequence"**.

Verified: `node --check` clean + a runtime test of the range / undersized / truncation / empty-write / segmented-codec logic (15/15).

finger563:
Addressed the remaining SDO/DS402 robustness comments (`f4a035449`):

- **Node pinning across multi-transaction actions** — `withSdo()` now takes an explicit node (default = field value). The **enable sequence** and the **identity read** snapshot the node once and thread it through every read/write, so editing the node field mid-action can no longer mix state/commands across drives. The enable sequence also **locks the node field** while running. *(covers the enable-sequence and identity comments)*
- **Enable sequence is cancellable** — **Quick stop** sets a cancel flag (and waits for the sequence to release the SDO channel) before issuing quick stop, checked after every awaited SDO op, so quick stop can no longer be undone by the sequence re-enabling the drive.
- **Fault reset** does both controlword writes (clear → `0x80`) **directly** instead of via `sendControlword()`, whose trailing `refreshDs402()` grabbed the SDO channel and failed the second write — bit 7 now gets its rising edge.
- **Object address validation** — `sdoUpload`/`sdoDownload` reject index outside `0x0000..0xFFFF` and subindex outside `0..255` before packing the low bits, so a typo like hex `16040` (`0x16040`) is no longer silently sent as `0x6040`. *(covers both read and write paths)*
- **Bounded uploads** — segmented uploads are capped at 64 KiB for both a declared-oversized total and a node that never sets the last-segment bit; on overflow the client sends an SDO abort and fails, so a faulty node can’t grow browser memory unbounded.
- **Serial EOF** — a clean `done:true` while still reading is now treated as a link failure (tears the UI down) instead of silently leaving it “Connected”; normal cancellation during `close()` is unaffected.
- **Mode-of-operation set** logs success only inside the successful write path (no success message after an abort/timeout/busy).

Verified: `node --check` clean + runtime tests of the address-validation and upload-bound logic.

PR #757, opened 2026-09-02
feat(canopen): configurable axis object-offset in Ds402Drive for multi-axis drives

finger563:
## Summary

`Ds402Drive` addressed a single axis at the standard device-profile indices (controlword `0x6040`, statusword `0x6041`, ...), so it could not drive the **second axis** of a multi-axis controller.

Basicmicro's MCP266, for example, mirrors every M1 object at `+0x800` for M2 (`0x6840`, `0x6841`, `0x687A`, ...). This adds `Config::object_offset`, applied via a new `axis_object()` helper.

## Details

- The offset is applied **only** to device-profile objects (`0x6000`-`0x6FFF`). Communication and identity objects (`< 0x6000`, e.g. `0x1000` device type, `0x1008` device name, `0x1018` identity) are shared by the whole device and are never offset.
- `object_offset` defaults to `0`, so existing single-axis behavior is unchanged.
- Two `Ds402Drive` instances (offset `0` and `0x800`) can share one `CanopenClient` to drive both channels of a dual-axis node.

```cpp
espp::Ds402Drive m1(client, {.object_offset = 0x000});
espp::Ds402Drive m2(client, {.object_offset = 0x800});
```

## Testing

- Existing canopen host test still passes (`ALL TESTS PASSED`).
- Exercised on real MCP266 hardware driving both channels' profile-position moves through two offset instances over one client.

### DIFF
--- a/components/canopen/include/detail/canopen_core.hpp
+++ b/components/canopen/include/detail/canopen_core.hpp
@@ -471,6 +471,40 @@ inline constexpr uint16_t OBJ_PROFILE_DECELERATION = 0x6084;       ///< Profile 
 inline constexpr uint16_t OBJ_TARGET_VELOCITY = 0x60FF;            ///< Target velocity (i32).
 /// @}
 
+/// @name Object-index bounds used for per-axis offsetting
+/// @{
+/// Inclusive low bound of the CiA 402 device-profile object range. Objects in
+/// [OBJ_DEVICE_PROFILE_MIN, OBJ_DEVICE_PROFILE_MAX] are the per-axis objects on
+/// a multi-axis drive.
+inline constexpr uint16_t OBJ_DEVICE_PROFILE_MIN = 0x6000;
+/// Inclusive high bound of the CiA 402 device-profile object range.
+inline constexpr uint16_t OBJ_DEVICE_PROFILE_MAX = 0x6FFF;
+/// Largest representable 16-bit CANopen object index.
+inline constexpr uint16_t OBJ_INDEX_MAX = 0xFFFF;
+/// Largest axis object-offset that keeps every device-profile index within the
+/// 16-bit index space (OBJ_INDEX_MAX - OBJ_DEVICE_PROFILE_MAX).
+inline constexpr uint16_t MAX_AXIS_OBJECT_OFFSET = OBJ_INDEX_MAX - OBJ_DEVICE_PROFILE_MAX;
+
+/// \brief Apply a per-axis object offset to a CiA 402 device-profile index.
+/// \details Only indices in the device-profile range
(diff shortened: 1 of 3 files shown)


finger563:
Addressed the review comment:

- **Offset range:** `axis_object()` now restricts the offset to the inclusive CiA 402 device-profile range `0x6000`–`0x6FFF`. Objects at `0x7000+` (e.g. manufacturer-specific ranges) are no longer offset, matching the documented behavior; comment clarified.

finger563:
Second round addressed:

- **Comment overclaim:** reworded — the helper no longer asserts that `>= 0x7000` objects are "shared / not axis-relative" (which can be false per device); it now just says those indices "are not offset by this helper and returned unchanged."
- **uint16_t wraparound:** the offset addition is done in `uint32_t`, and the offset result is only used when it still fits in `uint16_t` (`<= 0xFFFF`); otherwise the original index is returned rather than silently wrapping.

finger563:
Self-reviewed the full diff — no further issues. Every CiA 402 device-profile object accessor is wrapped with `axis_object()` while the identity/communication objects (0x1000/0x1001/0x1008/0x1018) are correctly left unoffset; the range is inclusive 0x6000-0x6FFF and the widened-add guards against a uint16_t wrap. Existing host test still passes.

finger563:
Addressed the latest round:

- **Magic numbers → named constants:** added `OBJ_DEVICE_PROFILE_MIN` (0x6000), `OBJ_DEVICE_PROFILE_MAX` (0x6FFF), `OBJ_INDEX_MAX` (0xFFFF) and `MAX_AXIS_OBJECT_OFFSET` (= `OBJ_INDEX_MAX - OBJ_DEVICE_PROFILE_MAX`) in `detail::ds402`, and the offset logic uses them.

- **Overflow no longer silently uses the wrong axis:** the offset is now validated once, in the `Ds402Drive` constructor. An `object_offset` greater than `MAX_AXIS_OBJECT_OFFSET` (i.e. one that could push a device-profile index past the 16-bit index space) is rejected with an error log and offsetting is disabled — rather than silently falling back per-call. With a valid offset, `axis_object()` can never overflow.

- **Host-testable + tested:** extracted the mapping into a pure `constexpr detail::ds402::apply_axis_offset(index, offset)` (which `Ds402Drive::axis_object()` now delegates to), and added `test_axis_offset` to the canopen host test covering (1) device-profile indices in 0x6000–0x6FFF are offset (incl. the inclusive boundaries), (2) indices outside the range (identity/comm objects, 0x5FFF, 0x7000) are not offset, and (3) an offset that would overflow returns the index unchanged. `ALL TESTS PASSED`.

PR #759, opened 2026-09-02
fix(ci): fix math / st7123touch / meshtastic example builds on ESP-IDF v6

finger563:
After the CI bump to ESP-IDF v6 (#751), three example builds fail. v6 ships **GCC 15.2** and compiles IDF’s own components with `-Wall -Wextra -Werror`, so warnings the older toolchain tolerated now break the build.

## Fixes
- **math** (`esp32`): the example applied `-Wdouble-promotion` **project-wide** (`add_compile_options` before `project()`), so it was also compiled into ESP-IDF — whose esp32 code (`esp_hw_support/port/esp32/rtc_clk.c`) trips it under the new GCC and fails IDF’s `-Werror` build. **Scope** the flag to the example’s own translation units via `target_compile_options(${COMPONENT_LIB} PRIVATE -Wdouble-promotion)` in `main/CMakeLists.txt`, keeping the float-only check without touching IDF. (The example is deliberately float-pure — all `0.05f` literals — so it still passes.)
- **st7123touch** (`esp32s3`) / **meshtastic** (`esp32s3`): `-Werror=missing-field-initializers` flagged the `std::function` `Config` members with **no default member initializer** (`St7123Touch::Config::write`/`read`, `MeshtasticNode::Config::transmit`) when a designated-initializer aggregate omits them. Give them a default `{nullptr}`, matching their sibling members (e.g. meshtastic’s `on_text{nullptr}`) — fixing it for every caller, not just these examples.

## Verification (IDF v6.0.1 / GCC 15.2, local)
- `math` (esp32) → **Project build complete**
- `st7123touch` (esp32s3) → **Project build complete**
- `meshtastic` (esp32s3): `transmit` was its **sole** missing-initializer error (identical pattern to st7123touch, which builds clean); it pulls in the large BSP/lvgl submodules, so it was not rebuilt locally.

### DIFF
--- a/components/math/example/CMakeLists.txt
+++ b/components/math/example/CMakeLists.txt
@@ -5,7 +5,11 @@ cmake_minimum_required(VERSION 3.20)
 set(ENV{IDF_COMPONENT_MANAGER} "0")
 include($ENV{IDF_PATH}/tools/cmake/project.cmake)
 
-add_compile_options(-Wdouble-promotion)
+# NOTE: -Wdouble-promotion (to keep this float-only example honest on targets
+# without hardware double support) is applied to the example's own code in
+# main/CMakeLists.txt, NOT globally: applying it project-wide also compiled it
+# into ESP-IDF's own components, whose code trips it under IDF v6's newer GCC
+# (e.g. esp_hw_support/port/esp32/rtc_clk.c) and fails their -Werror build.
 
 # add the component directories that we want to use
 set(EXTRA_COMPONENT_DIRS
--- a/components/math/example/main/CMakeLists.txt
+++ b/components/math/example/main/CMakeLists.txt
@@ -1,2 +1,6 @@
 idf_component_register(SRC_DIRS "."
                        INCLUDE_DIRS ".")
+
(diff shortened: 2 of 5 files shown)


finger563:
Follow-up: **meshtastic had a second v6 break** beyond the `transmit` initializer (the build stops at the first error, so it was hidden). Now fully built locally.

- **mbedtls 4.x** (`fb0475655`): ESP-IDF v6 ships mbedtls 4.x, which removed the public `<mbedtls/aes.h>` (the low-level AES API moved to `<mbedtls/private/aes.h>` as PSA Crypto became the public interface). `meshtastic_crypto.cpp` uses `mbedtls_aes_crypt_ctr` for the Meshtastic AES-256-CTR cipher, so it failed with `mbedtls/aes.h: No such file`. Selected the header by `MBEDTLS_VERSION_MAJOR` so it builds on both mbedtls 3.x (IDF v5) and 4.x (IDF v6); the `mbedtls_aes_*` calls are unchanged.

**meshtastic (esp32s3) now builds clean on IDF v6.0.1** — so all three examples in this PR are verified end-to-end.

PR #761, opened 2026-09-03
feat(mcp266): WebUSB / Web Serial console example + web app

finger563:
Adds a **companion** `components/mcp266/webapp_example` (the scripted `example` is unchanged) that turns an ESP32-S3 into a browser front-end for a Basicmicro MCP266: **configure**, **command**, and **view live status** of both motor channels over WebUSB / Web Serial.

Unlike the [CAN bridge](../canopen/can_bridge_example) (raw CAN + in-browser CANopen), this runs the `espp::Mcp266` driver **on the device** behind a small high-level protocol (`stream_frame` + `Dispatcher`, **module id 6**), so the web app needs zero CANopen/DS402 knowledge.

## Firmware
- `Twai` + `CanopenClient` + `Mcp266` + `UsbDevice` (vendor WebUSB + CDC Web Serial). A per-transport `Dispatcher` routes module-6 requests to a **worker task** (so blocking SDO never runs in the TinyUSB callback), and **all `Mcp266` SDO access is serialized under one mutex** (command handler + status streamer share the single SDO channel; the Twai RX task feeds `process_frame` from a different task, per the client contract).
- Protocol: `START` / `RESET_FAULTS` / `RESET_ESTOP` / `CONFIGURE_POSITION_LOOP` / `SET_POSITION_LIMITS` / `MOVE_TO_POSITION` / `DRIVE_SPEED` / `DRIVE_DUTY` / `GET_STATUS` / `SET_STATUS_STREAM` / `GET_DEVICE_INFO`; replies `OK` / `ERROR` / `STATUS` / `DEVICE_INFO`. `STATUS` streams per-axis position/velocity/statusword + battery/temperature.

## Web app
`components/mcp266/web/mcp266_console.html` — single-file, offline, WebUSB/Web Serial, reusing the DS402 panel’s transport + `stream_frame` framing. Per-axis (M1/M2) cards with **DS402 state decode + target-reached**, **configure** (position-loop clamp + fallback P, CiA 402 software limits, reset faults/e-stop, start node), **command** (profile-position move with vel/accel/decel; the inert speed/duty mirrors), **device telemetry**, and a device-side **live-status stream** toggle. Auto-listed in the apps index.

## Build / CI
CMakeLists mirror the `can_bridge_example` dual-mode setup (manager-on for registry users; manager-off + vendored `esp_tinyusb`/`tinyusb` submodules for CI). `build.yml` runs it on **esp32s3** with `IDF_COMPONENT_MANAGER=0`; manifest lists the new example.

## Verified (IDF v6.0.1 / GCC 15.2)
- Firmware builds clean (esp32s3, manager-off, no `managed_components`).
- Web app `node --check` clean.
- Firmware↔web-app **protocol round-trip test** (25-byte STATUS layout + move/configure command encode/decode, incl. negative positions) — **8/8**.
- Hardware test against a real MCP266 still pending (no device on hand).

Follow-up context: the separate design review flagged that surfacing `is_target_reached`/`get_state` on `Mcp266` would let the firmware avoid decoding the raw statusword here — not needed for this PR (the web app decodes DS402 state itself, as the DS402 panel does).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,6 +79,9 @@ jobs:
           target: esp32
         - path: 'components/mcp266/example'
           target: esp32
+        - path: 'components/mcp266/webapp_example'
+          target: esp32s3
+          command: 'IDF_COMPONENT_MANAGER=0 idf.py build'
         - path: 'components/bdc_driver/example'
           target: esp32s3
         - path: 'components/binary-log/example'
--- a/components/mcp266/idf_component.yml
+++ b/components/mcp266/idf_component.yml
@@ -8,6 +8,7 @@ maintainers:
 documentation: "https://esp-cpp.github.io/espp/motorcontrol/mcp266.html"
 examples:
   - path: example
+  - path: webapp_example
 tags:
   - cpp
   - Component
--- a/components/mcp266/web/mcp266_console.html
+++ b/components/mcp266/web/mcp266_console.html
@@ -1,0 +1,477 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>espp MCP266 Console (WebUSB / Web Serial)</title>
+  <meta name="description" content="Configure, command, and monitor a Basicmicro MCP266 dual-channel motor controller over a USB-to-CANopen bridge, from the browser: per-axis position loop setup, profile-position moves, and live status.">
+  <!--
+    espp MCP266 Console
+    ===================
+    A single-file, offline, dependency-free browser front-end for the espp MCP266
+    webapp example (components/mcp266/webapp_example). The ESP32 runs the
+    espp::Mcp266 driver and exposes a small HIGH-LEVEL protocol over USB; this
+    page speaks it over EITHER WebUSB (vendor 0xFF interface) OR Web Serial. All
+    CANopen/DS402 work happens on the device - this page just sends configure /
+    command / status messages.
+
+    Framed with the espp stream_frame v2 codec, routed on module id 6. Requests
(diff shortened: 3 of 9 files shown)


finger563:
Addressed the review comments in `7c97526ab`:

- **Axis not validated** (`:243`) — added `need_axis()` used by all axis-addressed requests (CONFIGURE / SET_LIMITS / MOVE / DRIVE_SPEED / DRIVE_DUTY); it rejects `pl[0] > 1` with `invalid_argument`, so an unexpected axis byte can no longer fall through `axis_of()` to M1 and command the wrong motor.
- **`active_transport` race** (`:339`/`:379`) — removed the store in the TinyUSB receive callback; the **RX worker is now the single writer**, setting it immediately before feeding each chunk to that chunk’s dispatcher. A frame arriving on the other endpoint between enqueue and dispatch can no longer retarget the reply/status for the frame being handled.
- **Meta description** (`web:7`) — reworded to match the architecture: the device runs the CANopen driver and exposes a high-level protocol over USB (WebUSB / Web Serial), not a "USB-to-CANopen bridge".

Verified: `node --check` clean, firmware rebuilds clean (esp32s3, manager-off, IDF v6.0.1).

finger563:
Addressed the remaining review comment in `7ca036ea6`:

- **`period_ms` unbounded** (`:306`) — a STATUS snapshot performs eight blocking SDO reads while holding the shared MCP mutex, so a small period could flood the CAN bus and starve command handling. `kSetStatusStream` now clamps the host-requested period to a documented **[50, 10000] ms** window (0 ⇒ the 200 ms default) via `std::clamp`, mirroring the streaming guard in `bldc_haptics_example.cpp`. The accepted range is documented on `kSetStatusStream` in `mcp266_protocol.hpp`. (The web console sends a fixed 250 ms, already inside the window.)

Verified: firmware rebuilds clean (esp32s3, manager-off, IDF v6.0.1).

PR #764, opened 2026-09-03
refactor(motor-control): shared MotorController concept + unified Axis-form API

finger563:
## Consistency pass — shared `MotorController` interface, unified Axis-form API

Third and final PR of the basicmicro / canopen / mcp266 review series. Stacked on
#763 (clarity); rebase onto `main` once #762 (bugs) and #763 (clarity) merge.

The review found that the two drivers modelling the **same** MCP236/266 hardware —
`espp::Basicmicro` (packet serial) and `espp::Mcp266` (CANopen) — presented
gratuitously different APIs for the same operations, and that `Basicmicro`'s
`drive_duty(m1, m2)` collided in meaning with `Mcp266`'s `drive_duty(axis, duty)`
(one meant "both channels", the other "one channel"). This unifies them.

### 1. Shared `espp::MotorController` concept + `espp::MotorAxis`
New `base_component/include/motor_controller.hpp`: a compile-time concept naming the
common dual-channel surface both drivers should present — `drive_duty` / `drive_speed`
by `MotorAxis`, `read_encoder` / `read_speed` (signed `int32`), `reset_estop`,
`read_main_battery_voltage`, `read_temperature`. Both classes `using Axis = MotorAxis`
and `static_assert` conformance, so generic code can drive either transport by axis
and the contract is enforced at compile time.

### 2. `Basicmicro`: Axis-form API (collapses ~200 lines of M1/M2 duplication)
Every per-channel command pair (`drive_m1_*`/`drive_m2_*`, `read_encoder_m1/m2`,
`read_encoder_speed_m1/m2`, `set/read_velocity_pid_m1/m2`, `set/read_position_pid_m1/m2`,
buffered variants) collapses to a single method taking `Axis`, selecting the command
byte from the axis.

- The **both-channels** commands are renamed `drive_both_duty` / `drive_both_speed` /
  `drive_both_speed_accel` / `buffered_drive_both_*` so the unqualified name always
  means "one channel" (matching `Mcp266`, killing the collision).
- `e_stop_reset` → `reset_estop`.
- `read_encoder` now returns a **signed `int32`** count (was `uint32`) — reverse
  quadrature reads negative, as callers expect; a status-byte overload is retained.

**Wire format is unchanged** — each Axis method emits the exact command byte + payload
the old M1/M2 method did.

### 3. `Mcp266` alignment tidy-ups (deferred from the clarity PR)
- `configure_position_loop`: `ec` moved to the **last** parameter (restoring the
  codebase convention) with a convenience overload using the default position P gain.
- `set_position_limits` → `set_software_position_limits` (disambiguates it from the
  manufacturer position-PID clamp).

### Verification
- `basicmicro` example builds clean on **IDF v6.0.1** (esp32); host test unchanged.
- `mcp266` example builds clean on **IDF v6.0.1**.
- Both `static_assert(MotorController<...>)` pass — the two drivers are now
  interchangeable through the concept.

### ⚠️ Breaking API changes (both drivers are pre-1.0 components)
- `Basicmicro`: `drive_m1_duty(d)` → `drive_duty(Axis::M1, d)`; `drive_duty(m1, m2)`
  → `drive_both_duty(m1, m2)`; same pattern for speed / PID / buffered / encoder reads;
  `read_encoder_m1(uint32_t&, ...)` → `read_encoder(Axis::M1, int32_t&, ...)`;
  `e_stop_reset` → `reset_estop`.
- `Mcp266`: `set_position_limits` → `set_software_position_limits`;
  `configure_position_loop(...)` gains a required position-P argument before `ec`
  (or use the 4-arg default-gain overload).

> Note: the mcp266 WebUSB example (#761) will need a one-line `configure_position_loop`
> call update once this lands; tracked with that PR.

### DIFF
--- a/.github/workflows/upload_components.yml
+++ b/.github/workflows/upload_components.yml
@@ -55,6 +55,7 @@ jobs:
             components/aw9523
             components/base_component
             components/base_peripheral
+            components/motor_controller
             components/basicmicro
             components/mcp266
             components/bdc_driver
--- a/components/base_component/include/motor_controller.hpp
+++ b/components/base_component/include/motor_controller.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+#include <concepts>
+#include <cstdint>
+#include <system_error>
+
+namespace espp {
+
+/// \brief Motor channel selector shared by the espp motor-controller drivers.
+/// \details M1 / M2 are the two output channels of a dual-channel controller
+///          (e.g. a Basicmicro MCP236/MCP266). Drivers that model this hardware
+///          family select a channel with this type so generic code can drive
+///          either transport (packet-serial espp::Basicmicro or CANopen
+///          espp::Mcp266) uniformly.
+enum class MotorAxis : uint8_t { M1 = 0, M2 = 1 };
+
+/// \brief Compile-time contract for a dual-channel motor controller.
+/// \details Both espp::Basicmicro (packet serial) and espp::Mcp266 (CANopen)
+///          model the same MCP236/266 hardware and satisfy this concept, so
+///          generic code can command either by axis. Every operation follows the
+///          espp convention: it returns \c true on success and sets \p ec on
+///          failure (leaving it cleared on success). Units are encoder counts
+///          (position), counts/s (velocity / "qpps"), a signed duty (±32767),
+///          volts, and degrees Celsius.
+/// \note Conformance is about the API surface, not that every command is active
+///       on every device in its current mode: a controller may accept a duty /
+///       speed command it cannot act on (e.g. Mcp266's manufacturer speed/duty
+///       mirror is inert on firmware where only CiA 402 position mode drives).
+///       Position control and per-device configuration are intentionally NOT
+///       part of this common surface -- they differ too much between transports.
+template <typename T>
+concept MotorController = requires(T t, MotorAxis axis, int16_t duty, int32_t qpps, int32_t count,
+                                   float value, std::error_code &ec) {
+  /// Open-loop signed duty on one channel.
+  { t.drive_duty(axis, duty, ec) } -> std::same_as<bool>;
+  /// Closed-loop signed speed (counts/s) on one channel.
+  { t.drive_speed(axis, qpps, ec) } -> std::same_as<bool>;
+  /// Read the signed encoder count of one channel.
+  { t.read_encoder(axis, count, ec) } -> std::same_as<bool>;
+  /// Read the signed encoder speed (counts/s) of one channel.
+  { t.read_speed(axis, qpps, ec) } -> std::same_as<bool>;
+  /// Clear a latched emergency stop.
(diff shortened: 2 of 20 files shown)


finger563:
⚠️ **Rebase note — #761 (mcp266 webapp) is now merged to `main`.** Two call sites in `components/mcp266/webapp_example/main/mcp266_webapp_example.cpp` use the *old* signatures this PR changes, and since this branch does not contain the webapp file, **git will not flag a conflict** — `main` would build-break when this lands. When rebasing this PR onto `main`, fold in:

```cpp
// :278  configure_position_loop — ec moves to LAST, fallback_p before it
mcp.configure_position_loop(axis_of(pl[0]), rd_i32(pl, 1), rd_i32(pl, 5), rd_i32(pl, 9), ec)

// :285  set_position_limits -> set_software_position_limits
mcp.set_software_position_limits(axis_of(pl[0]), rd_i32(pl, 1), rd_i32(pl, 5), ec)
```

No protocol/wire change — same payload byte offsets, just the C++ argument order / method name.

finger563:
Addressed the latest review in `c5ddd73be`:

- **Out-of-range axis silently targets M2** (`basicmicro.hpp`, `mcp266.hpp`) — added a `check_axis()` guard that every public Axis-form method now calls before any I/O; it rejects anything but `M1`/`M2` with `invalid_argument`. This also covers `Mcp266::axis_state` (all its callers validate first), so a `MotorAxis` decoded from an untrusted byte can no longer fall through the `? :` dispatch to M2. Verified: basicmicro + mcp266 examples build clean (IDF v6.0.1).

- **`motor_controller` packaging** — added `components/motor_controller/README.md`. On the example + `build.yml` matrix request: `motor_controller` is a header-only interface/types component, exactly like `bldc_types`, which ships **without** an example, `examples:` manifest entry, or build matrix (and is published fine). A concept has nothing runnable to demonstrate beyond the drivers that implement it, and those are already exercised by the `basicmicro`/`mcp266` examples in CI — so I added the README but intentionally did not add a placeholder example, to match the existing precedent. Happy to add one if you would prefer to diverge from how `bldc_types` is packaged.

finger563:
Follow-ups addressed in `331dbf3f8`:

- **`MotorAxis` formatter printed `M2` for invalid values** — now an explicit `switch` that formats an out-of-range value as `UNKNOWN`, so diagnostics never mislabel malformed input as a valid channel. (Verified with a host test: `M1` / `M2` / `UNKNOWN`, clean under `-Wall -Wextra`.)
- **`Mcp266::drive(Axis)` bypassed the guard** — the advanced-access accessor now returns `Ds402Drive*` and yields `nullptr` for anything but `M1`/`M2`, so a decoded-byte axis can no longer reach the M2 drive through it. No in-tree callers; pre-1.0, so the reference→pointer change is safe. mcp266 example builds clean (IDF v6.0.1).

PR #780, opened 2026-09-07
feat(cobs): Fixed issue when exactly 255 bytes (non zero) are encoded.

jschuld:
## Description
When encoding data of exactly 255 bytes in length, where the data is all non-zero, the last byte encoded is missing. This results in the encoded data not being valid COBS and can't be decoded.

## Motivation and Context
While this is an edge case, it is a valid case and breaks any COBS decoding. This issue was identified when diagnosing issues where COBS decoding would fail intermittedly. A different COBS library was de decoding party. Initial suspicion was actual corruption of data on the line. However, after digging into this the issue was identified to be the COBS encoding for very specific cases (the above case)

## How has this been tested?
- Created a unittest (included) which reproduced the issue.
- Ran fix on actual hardware, resulting in zero problems after the fix. 

N/A

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation Update
- [ ] Software change

## Checklist:

### Software

- [x] I have added tests to cover my changes.
- [-] (N/A Test added to exising tests) I have updated the `.github/workflows/build.yml` file to add my new test to the automated cloud build github action.

### DIFF
--- a/components/cobs/example/main/cobs_example.cpp
+++ b/components/cobs/example/main/cobs_example.cpp
@@ -682,6 +682,24 @@ void test_edge_cases(espp::Logger &logger) {
           decoded.size(), large_packet.size(), encoded.size());
     }
   }
+
+  // Test 5: Packet of exactly block len with no zero's in the data
+  {
+    std::vector<uint8_t> block_len_packet(255, 0x42);
+
+    std::vector<uint8_t> encoded = Cobs::encode_packet(block_len_packet);
+    std::vector<uint8_t> decoded = Cobs::decode_packet(encoded);
+
+    bool success = (decoded.size() == block_len_packet.size()) &&
+                   (std::memcmp(decoded.data(), block_len_packet.data(), block_len_packet.size()) == 0);
+    if (success) {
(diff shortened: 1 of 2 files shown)


components/cobs/example/main/cobs_example.cpp
jschuld, line 610: Note: This tests seems to attempt to test the same as the test that is added. However, I am unsure if there is a specific reason that 254 is chosen, so I have opted to add an extra test.

components/cobs/src/cobs.cpp
```diff
@@ -44,8 +44,7 @@ size_t Cobs::encode_packet(std::span<const uint8_t> data, std::span<uint8_t> out
       *codep = code;
       code = 1;
       codep = encode;
-      if (!byte)
-        ++encode;
+      ++encode;
```
jschuld, line 47: If the loop ends because of no more data, we still want to ready the next encoding position for the delimiter to be added.

finger563:
Thanks for the fix!

